PR #4, opened 2020-02-13
Require bootstrap.php if exists, to load all necessary .env files

Naktibalda:
Fixes https://github.com/Codeception/Codeception/issues/5411

Must be released as 2.0.0

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -202,6 +202,13 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         }
 
         $this->kernel = new $this->kernelClass($this->config['environment'], $this->config['debug']);
+
+        $bootstrapFile = $this->kernel->getProjectDir() . '/config/bootstrap.php';
+        if (file_exists($bootstrapFile)) {
+            $_ENV['APP_ENV'] = $this->config['environment'];
+            require $bootstrapFile;
+        }
+
         $this->kernel->boot();
 
         if ($this->config['cache_router'] === true) {


c33s:
@TavoNiievez  via slack:

> How does #4 need to be updated according to [vworldat's][1] comment on Nov 29?

a solution can be that we use the kernel constant to handle the dotenv loading correctly between the different symfony versions.

something like:
```php
    if (Kernel::VERSION_ID < 50000) {
        // load bootstrap file
    } else if (...) {
        // just instance the dotenv component
    }
```


most current version (symfony 5.2) to boot up dotenv 
`bin/console`:
```php
if (!class_exists(Application::class) || !class_exists(Dotenv::class)) {
    throw new LogicException('You need to add "symfony/framework-bundle" and "symfony/dotenv" as Composer dependencies.');
}

$input = new ArgvInput();
if (null !== $env = $input->getParameterOption(['--env', '-e'], null, true)) {
    putenv('APP_ENV='.$_SERVER['APP_ENV'] = $_ENV['APP_ENV'] = $env);
}

if ($input->hasParameterOption('--no-debug', true)) {
    putenv('APP_DEBUG='.$_SERVER['APP_DEBUG'] = $_ENV['APP_DEBUG'] = '0');
}

(new Dotenv())->bootEnv(dirname(__DIR__).'/.env');

if ($_SERVER['APP_DEBUG']) {
    umask(0000);

    if (class_exists(Debug::class)) {
        Debug::enable();
    }
}

$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
$application = new Application($kernel);
$application->run($input);
```

so we will end up to something like
```php
    //should already be autoloaded here i assume
    if (!class_exists(Dotenv::class)) {
        throw new LogicException('You need to add "symfony/dotenv" as Composer dependencies.');
    }
    
    putenv('APP_ENV='.$_SERVER['APP_ENV'] = $_ENV['APP_ENV'] = $this->config['environment']); //$this->config['environment'] should have `test` as default
    
    //maybe also add a `debug` config setting

    //read the kernel version but allow the user to override it via config. of course we should use better names (constants) than `1`,`2` and `3`. default should be `0` to use the kernel version by default.
    //don't checked which version did the change. i think it was 5.0 but we have to check the dotenv package
    if (Kernel::VERSION_ID >= 50000 || $config[env_handling] == '3') {
        //https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.1/public/index.php#L10
        //https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.2/public/index.php#L10
        
        (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
    } else if ((Kernel::VERSION_ID < 50000 && file_exists) || $config[env_handling] == '2') {
        //check if file exists. the bootstrap.php file is only generated by flex on new projects. so it's possible to have symfony 3.4 projects with the old loading behavior
        //https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.4/config/bootstrap.php //same as 3.3
        //https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/bootstrap.php
        //https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.4/config/bootstrap.php

        require dirname(__DIR__).'/config/bootstrap.php';
    } else if (Kernel::VERSION_ID <= 30400 || $config[env_handling] == '1') {
        //older symfony versions loaded their data via parameters and prefixed environment variables
        //https://github.com/symfony/symfony-standard/blob/3.4/bin/console
    } else {
        //exception
    }
    
    // looks like the kernel should be booted after the environment is set
    $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
```

@Naktibalda

> Must be released as 2.0.0

can't we put it into 1.x with the same behavior then before and do a deprecated warning. with 2.x we change the default behavior and remove the deprecated warning.

references:
- https://symfony.com/doc/current/configuration/dot-env-changes.html#updating-my-application
- https://symfony.com/blog/improvements-to-the-handling-of-env-files-for-all-symfony-versions
- https://github.com/symfony/dotenv/blob/5.x/CHANGELOG.md

[1]: https://github.com/Codeception/module-symfony/issues/99#issuecomment-735415461

c33s:
after i posted i had an even better idea. the `Kernel::VERSION_ID` might not be the best indication of which version of dotenv is installed. we should simply query for the installed dotenv version.

https://github.com/symfony/dotenv/blob/5.x/CHANGELOG.md

https://packagist.org/packages/composer/package-versions-deprecated (we should use the deprecated version of this package as the version ocramius requires to high php versions most of the time.
```
$version = \PackageVersions\Versions::getVersion('symfony/dotenv');
```
then parse it and comare it with `composer/semver` https://packagist.org/packages/composer/semver against the versions in the changelog.md + some custom handling for versions which could have an upgraded directory structure

TavoNiievez:
@Naktibalda I'm ready to release version `2.0`, allowing the installation of Codeception 5 when it is released may be a minor update `2.0.x`.

I never had the issue that this PR solves, maybe because I added `.env.test` to my `codeception.yml` file

If this requires a lot more work and is a BC it can be redirected to Milestone `3.0`.

TavoNiievez:
Replaced by #190.

PR #5, opened 2020-02-20
If symfony is install on another project/dir than codeception, this ensure that its autoloader is load too

XuruDragon:
If symfony is install on another project/dir than codeception, this ensure that its autoloader is load too


Naktibalda:
Having two vendor directories is recipe for disaster. Codeception comes with its own copy of a few Symfony components, so eventually you will have issues caused by incompatible libraries in different vendor directories.

PR #10, opened 2020-09-05
Refactoring work

TavoNiievez:
closes Codeception/Codeception#5699

 - [x] Added new function: `amLoggedInAs` to login a user. 

<details>
  <summary>Click to see added function.</summary>

```php
    public function amLoggedInAs(UserInterface $user, $firewallName = 'main', $firewallContext = null)
    {
        $container = $this->_getContainer();
        if (!$container->has('session')) {
            return;
        }

        $session = $this->grabService('session');

        if ($this->config['guard']) {
            $token = new PostAuthenticationGuardToken($user, $firewallName, $user->getRoles());
        } else {
            $token = new UsernamePasswordToken($user, null, $firewallName, $user->getRoles());
        }

        if ($firewallContext) {
            $session->set('_security_'.$firewallContext, serialize($token));
        } else {
            $session->set('_security_'.$firewallName, serialize($token));
        }

        $session->save();

        $cookie = new Cookie($session->getName(), $session->getId());
        $this->client->getCookieJar()->set($cookie);
    }
```

</details>

```php
        $user = $I->grabEntityFromRepository(User::class, ['email' => 'john_doe@gmail.com']);
        $I->amLoggedInAs($user);
```

- [x] Added [guard](https://symfony.com/components/Guard) config, in case the authentication system is based on it.
```diff
    public $config = [
        'app_path' => 'app',
        'var_path' => 'app',
        'kernel_class' => null,
        'environment' => 'test',
        'debug' => true,
        'cache_router' => false,
        'em_service' => 'doctrine.orm.entity_manager',
        'rebootable_client' => true,
        'mailer' => self::SWIFTMAILER,
+        'guard' => false
    ];
```

- [x] Added `SymfonyModuleInterface` to improve the DX.

<details>
  <summary>Click to see added interface.</summary>

```php
interface SymfonyModuleInterface extends DoctrineProvider, PartedModule
{
    public function _after(TestInterface $test);

    public function _before(TestInterface $test);

    public function _getContainer();

    public function _getEntityManager();

    public function _initialize();

    public function _parts();

    public function amOnRoute($routeName, array $params = []);

    public function dontSeeEmailIsSent();

    public function grabService($service);

    public function invalidateCachedRouter();

    public function amLoggedInAs(UserInterface $user, $firewallName = 'main', $firewallContext = null);

    public function onReconfigure($settings = []);

    public function persistService($serviceName, $isPermanent = false);

    public function rebootClientKernel();

    public function runSymfonyConsoleCommand($command, $parameters = [], $consoleInputs = [], $expectedExitCode = 0);

    public function seeCurrentRouteIs($routeName, array $params = []);

    public function seeEmailIsSent($expectedCount = null);

    public function seeInCurrentRoute($routeName);

    public function unpersistService($serviceName);
}
```

</details>

- [x] Updated dependencies (as far as possible).
- [x] Updated docs.
- [x] Methods and functions were ordered.
- [x] multiple minor visual improvements in documentation.
- [x] Improved compatibility logic:
```diff
- //Alias for Symfony < 4.3
- if (!class_exists('Symfony\Component\HttpKernel\HttpKernelBrowser') && class_exists('Symfony\Component\HttpKernel\Client')) {
+ if (Kernel::VERSION_ID < 40300) {
-    class_alias('Symfony\Component\HttpKernel\Client', 'Symfony\Component\HttpKernel\HttpKernelBrowser');
+    class_alias(Client::class, HttpKernelBrowser::class);
}
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2011 Michael Bodnarchuk and contributors
+Copyright (c) 2011-2020 Michael Bodnarchuk and contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
     "require": {
         "php": ">=5.6.0 <8.0",
         "codeception/lib-innerbrowser": "^1.3",
-        "codeception/codeception": "^4.0",
-        "symfony/symfony": ">=2.8 <5.2"
+        "codeception/codeception": "^4.0"
     },
     "require-dev": {
         "codeception/util-robohelpers": "dev-master",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name":"codeception/module-symfony",
     "description":"Codeception module for Symfony framework",
     "keywords":["codeception", "symfony"],
-    "homepage":"http://codeception.com/",
+    "homepage":"https://codeception.com/",
     "type":"library",
     "license":"MIT",
     "authors":[
@@ -13,17 +13,18 @@
     "minimum-stability": "RC",
     "require": {
         "php": ">=5.6.0 <8.0",
-        "codeception/lib-innerbrowser": "^1.0",
-        "codeception/codeception": "^4.0"
+        "codeception/lib-innerbrowser": "^1.3",
+        "codeception/codeception": "^4.0",
+        "symfony/symfony": ">=2.8 <5.2"
     },
     "require-dev": {
         "codeception/util-robohelpers": "dev-master",
-        "codeception/module-asserts": "dev-master | ^1.0",
-        "codeception/module-doctrine2": "dev-master | ^1.0",
-        "codeception/module-phpbrowser": "dev-master | ^1.0",
-        "codeception/module-rest": "dev-master | ^1.0",
-        "codeception/module-sequence": "dev-master | ^1.0",
-        "vlucas/phpdotenv": "^3.6"
+        "codeception/module-asserts": "^1.3",
+        "codeception/module-doctrine2": "^1.0",
+        "codeception/module-phpbrowser": "^1.0",
+        "codeception/module-rest": "^1.2",
+        "codeception/module-sequence": "^1.0",
+        "vlucas/phpdotenv": "^4.1"
     },
     "autoload":{
         "classmap": ["src/"]
--- a/documentation.md
+++ b/documentation.md
@@ -21,12 +21,13 @@ This module uses Symfony Crawler and HttpKernel to emulate requests and test res
 
 #### Example (`functional.suite.yml`) - Symfony 4 Directory Structure
 
-    modules:
-       enabled:
-          - Symfony:
-              app_path: 'src'
-              environment: 'test'
-
+```yaml
+modules:
+   enabled:
+      - Symfony:
+          app_path: 'src'
+          environment: 'test'
+```
 
 ### Symfony 3.x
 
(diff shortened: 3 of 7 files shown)


c33s:
thanks for your time and effort to make the symfony module better. sadly i am quite unhappy with this PR. 

my first impression: too many changes for adding a simple function like `amLoggedInAs`. you added 656 and removed 512 lines. you have refactored half of the module in one big PR. this is quite hard to review. 

you require higher versions from different packages for what exactly? why the symfony module interface? ... for me style changes must not be in the same PR as functional changes.
each of this changes should be its own PR where things can be discussed and be included or rejected one by one. as i use the symfony module quite a lot i want to know why something needs to be changed or why a thing should be refactored. there are no tests for the changes. 
if everything is in one file like in this case (in general i would vote for using more traits) add new functions to the bottom, a diff showing me the follwing

2020-09-05 16_59_59- #  Added new function_ 'amLoggedInAs' to login a user (and more)  by TavoNiieve


even if you haven't touched the email system is really hard to review.

i miss the option to define roles. i first have to create a/the user entity with it's correct roles, persist it, fetch it only to provide it to the logged in method. instead of 
```
$I->amLoggedInAs('i_choose_any_username_in_the_test', 'main', ['ROLE_ADMIN', 'ROLE_ALLOW_X'])
``` 
what if you use basic auth, then there is no user entity, i have to mock one to log in. feels really unhandy for me or do i overlook something?


the PR, as it is now, is a nogo for me. so it's 👎 from my side.

TavoNiievez:
@c33s I just followed the [Symfony documentation for this type of authentication](https://symfony.com/doc/current/testing/http_authentication.html), if you try to create a token with a `UserInterface` that is not persisted (and does not have an identifier) ​​it will result in an Exception.

Furthermore, creating and authenticating a user with a simple string entails many problems, for example if I want to test behaviors related to the currently logged in user and the user does not have, for example,  associated invoices since the user was created on the go with a string... i would simply have to do exactly what you suggest is unnecessary, create a user, add invoices and roles to it, persist it, and finally get it from the session.

Now, you are **_supposed_** to be able to create `TestFixtures` and run them before each test.

Next, _there are_ tests for this module, you can create some and send them to [Codeception/symfony-demo](https://github.com/Codeception/symfony-demo), and [wait for them to be merged](https://github.com/Codeception/symfony-demo/pull/12).

I fully agree that these should be individual PRs, so feel free to submit another PR with just one of the changes described here.

and finally, i will change the title so as not to have to explain why the libraries of a module need to be updated

c33s:
@TavoNiievez 

> I just followed the Symfony documentation for this type of authentication, if you try to create a token with a UserInterface that is not persisted (and does not have an identifier) ​​it will result in an Exception.

you followed the `testing` version (5.1) have you tested that it is at least compatible with all current supproted symfony versions? don't know when i used my code the last time but as far as i can remember the code was compatible with 3.4 and 4.4 without an exception. which symfony version do you use?

for handy testing of the backend permissions i don't care about the user i simply want to login and have the correct roles (can a admin delete a user, is the delete button hidden for an editor,...) so most of the times i don't need a user object.

for the case you described the manual user creation is part of the test. so having both options is better than only have one.

> I fully agree that these should be individual PRs, so feel free to submit another PR with just one of the changes described here.

well as i have not written it, it's much harder for me to split out your changes instead of you submitting separate PR's.

> and finally, i will change the title so as not to have to explain why the libraries of a module need to be updated

changing the title does not make the quality of this PR better :)

TavoNiievez:
@c33s We seem to have reached an impasse, I just don't see the need to be so inflammatory.
neither am I going to divide, nor are you going to send.
Anyway I will leave this as is so that someone knows how to send PR does it, and that obviously it is not you since it seems to be too much work.

PR #11, opened 2020-09-05
Added amLoggedInAs function (and more)

ghost:
Related to [Codeception issue 5699](https://github.com/Codeception/Codeception/issues/5699)

 - [x] Added new function: `amLoggedInAs` to login a user. 

<details>
  <summary>Click to see added function.</summary>

```php
    public function amLoggedInAs(UserInterface $user, $firewallName = 'main', $firewallContext = null)
    {
        $container = $this->_getContainer();
        if (!$container->has('session')) {
            $this->fail("Symfony container doesn't have 'session' service");
            return;
        }

        $session = $this->grabService('session');

        if ($this->config['guard']) {
            $token = new PostAuthenticationGuardToken($user, $firewallName, $user->getRoles());
        } else {
            $token = new UsernamePasswordToken($user, null, $firewallName, $user->getRoles());
        }

        if ($firewallContext) {
            $session->set('_security_'.$firewallContext, serialize($token));
        } else {
            $session->set('_security_'.$firewallName, serialize($token));
        }

        $session->save();

        $cookie = new Cookie($session->getName(), $session->getId());
        $this->client->getCookieJar()->set($cookie);
    }
```

</details>

```php
        $user = $I->grabEntityFromRepository(User::class, ['email' => 'john_doe@gmail.com']);
        $I->amLoggedInAs($user);
```

- [x] Added [guard](https://symfony.com/components/Guard) config, in case the authentication system is based on it.
```diff
    public $config = [
        'app_path' => 'app',
        'var_path' => 'app',
        'kernel_class' => null,
        'environment' => 'test',
        'debug' => true,
        'cache_router' => false,
        'em_service' => 'doctrine.orm.entity_manager',
        'rebootable_client' => true,
        'mailer' => self::SWIFTMAILER,
+        'guard' => false
    ];
```

- [x] Updated dependencies (as far as possible).
- [x] Updated docs.
- [x] Improved compatibility logic:
```diff
- //Alias for Symfony < 4.3
- if (!class_exists('Symfony\Component\HttpKernel\HttpKernelBrowser') && class_exists('Symfony\Component\HttpKernel\Client')) {
+ if (Kernel::VERSION_ID < 40300) {
    class_alias('Symfony\Component\HttpKernel\Client', 'Symfony\Component\HttpKernel\HttpKernelBrowser');
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "codeception/module-phpbrowser": "^1.0",
         "codeception/module-rest": "^1.2",
         "codeception/module-sequence": "^1.0",
-        "vlucas/phpdotenv": "^4.1"
+        "vlucas/phpdotenv": "^3.6 | ^4.1"
     },
     "autoload":{
         "classmap": ["src/"]
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name":"codeception/module-symfony",
     "description":"Codeception module for Symfony framework",
     "keywords":["codeception", "symfony"],
-    "homepage":"http://codeception.com/",
+    "homepage":"https://codeception.com/",
     "type":"library",
     "license":"MIT",
     "authors":[
@@ -13,17 +13,17 @@
     "minimum-stability": "RC",
     "require": {
         "php": ">=5.6.0 <8.0",
-        "codeception/lib-innerbrowser": "^1.0",
+        "codeception/lib-innerbrowser": "^1.3",
         "codeception/codeception": "^4.0"
     },
     "require-dev": {
         "codeception/util-robohelpers": "dev-master",
-        "codeception/module-asserts": "dev-master | ^1.0",
-        "codeception/module-doctrine2": "dev-master | ^1.0",
-        "codeception/module-phpbrowser": "dev-master | ^1.0",
-        "codeception/module-rest": "dev-master | ^1.0",
-        "codeception/module-sequence": "dev-master | ^1.0",
-        "vlucas/phpdotenv": "^3.6"
+        "codeception/module-asserts": "^1.3",
+        "codeception/module-doctrine2": "^1.0",
+        "codeception/module-phpbrowser": "^1.0",
+        "codeception/module-rest": "^1.2",
+        "codeception/module-sequence": "^1.0",
+        "vlucas/phpdotenv": "^4.1"
     },
     "autoload":{
         "classmap": ["src/"]
--- a/documentation.md
+++ b/documentation.md
@@ -1,24 +1,4 @@
 # Symfony
-## Installation
-
-If you use Codeception installed using composer, install this module with the following command:
-
-```
-composer require --dev codeception/module-symfony
-```
-
-Alternatively, you can enable `Symfony` module in suite configuration file and run
- 
-```
-codecept init upgrade4
-```
-
-This module was bundled with Codeception 2 and 3, but since version 4 it is necessary to install it separately.   
-Some modules are bundled with PHAR files.  
-Warning. Using PHAR file and composer in the same project can cause unexpected errors.  
-
-## Description
-
 
 
 This module uses Symfony Crawler and HttpKernel to emulate requests and test response.
--- a/documentation.md
+++ b/documentation.md
@@ -235,6 +235,23 @@ $I->amOnPage('/register');
  * `param string` $page
 
 
+### amLoggedInAs
+
+Login with the given UserInterface.
+
+```php
+<?php
+$user = $I->grabEntityFromRepository(User::class, [
+    'email' => 'john_doe@gmail.com'
+]);
+$I->amLoggedInAs($user);
+```
+
+ * `param UserInterface` $user
+ * `param string` $firewallName
+ * `param null` $firewallContext
+
+
 ### amOnRoute
  
 Opens web page using route name and parameters.
--- a/documentation.md
+++ b/documentation.md
@@ -1,4 +1,24 @@
 # Symfony
+## Installation
+
+If you use Codeception installed using composer, install this module with the following command:
+
+```
+composer require --dev codeception/module-symfony
+```
+
+Alternatively, you can enable `Symfony` module in suite configuration file and run
+ 
+```
+codecept init upgrade4
+```
+
+This module was bundled with Codeception 2 and 3, but since version 4 it is necessary to install it separately.   
+Some modules are bundled with PHAR files.  
+Warning. Using PHAR file and composer in the same project can cause unexpected errors.  
+
+## Description
+
 
 
 This module uses Symfony Crawler and HttpKernel to emulate requests and test response.
@@ -18,6 +38,8 @@ This module uses Symfony Crawler and HttpKernel to emulate requests and test res
 * debug: true - turn on/off debug mode
 * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
 * rebootable_client: 'true' - reboot client's kernel before each request
+* mailer: 'symfony_mailer' - choose the mailer used by your application
+* guard: 'false' - Specifies if Guard component is used as authentication system
 
 #### Example (`functional.suite.yml`) - Symfony 4 Directory Structure
 
@@ -38,6 +60,8 @@ This module uses Symfony Crawler and HttpKernel to emulate requests and test res
 * debug: true - turn on/off debug mode
 * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
 * rebootable_client: 'true' - reboot client's kernel before each request
+* mailer: 'swiftmailer' - choose the mailer used by your application
+* guard: 'false' - Specifies if Guard component is used as authentication system
 
 #### Example (`functional.suite.yml`) - Symfony 3 Directory Structure
 
@@ -58,6 +82,8 @@ This module uses Symfony Crawler and HttpKernel to emulate requests and test res
 * em_service: 'doctrine.orm.entity_manager' - use the stated EntityManager to pair with Doctrine Module.
 * cache_router: 'false' - enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
 * rebootable_client: 'true' - reboot client's kernel before each request
+* mailer: 'swiftmailer' - choose the mailer used by your application
+* guard: 'false' - Specifies if Guard component is used as authentication system
 
 ### Example (`functional.suite.yml`) - Symfony 2.x Directory Structure
 
@@ -220,6 +246,24 @@ Authenticates user for HTTP_AUTH
  * `param` $password
 
 
+### amLoggedInAs
+ 
+Login with the given UserInterface.
+That UserInterface object must have a persistent identifier.
+
+```php
+<?php
+$user = $I->grabEntityFromRepository(User::class, [
+'email' => 'john_doe@gmail.com'
+]);
+$I->amLoggedInAs($user);
+```
+
+ * `param UserInterface` $user
+ * `param string` $firewallName
+ * `param null` $firewallContext
+
+
 ### amOnPage
  
 Opens the page for the given relative URI.
@@ -233,23 +277,6 @@ $I->amOnPage('/register');
 ```
 
  * `param string` $page
-
-
-### amLoggedInAs
-
-Login with the given UserInterface.
-
-```php
-<?php
-$user = $I->grabEntityFromRepository(User::class, [
-    'email' => 'john_doe@gmail.com'
-]);
-$I->amLoggedInAs($user);
-```
-
- * `param UserInterface` $user
- * `param string` $firewallName
- * `param null` $firewallContext
 
 
 ### amOnRoute
--- a/src/Codeception/Lib/Connector/Symfony.php
+++ b/src/Codeception/Lib/Connector/Symfony.php
@@ -2,9 +2,9 @@
 namespace Codeception\Lib\Connector;
 
 use Symfony\Component\HttpKernel\HttpKernelBrowser;
+use Symfony\Component\HttpKernel\Kernel;
 
-//Alias for Symfony < 4.3
-if (!class_exists('Symfony\Component\HttpKernel\HttpKernelBrowser') && class_exists('Symfony\Component\HttpKernel\Client')) {
+if (Kernel::VERSION_ID < 40300) {
     class_alias('Symfony\Component\HttpKernel\Client', 'Symfony\Component\HttpKernel\HttpKernelBrowser');
 }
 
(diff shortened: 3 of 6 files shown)


src/Codeception/Module/SymfonyModuleInterface.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function _before(TestInterface $test);
+
+    public function _getContainer();
+
+    public function _getEntityManager();
+
+    public function _initialize();
+
+    public function _parts();
```
Naktibalda, line 20: What's the purpose of this interface? Is it documentation?

* Standard hooks that are common to all modules are defined in parent class [Codeception\Module](https://github.com/Codeception/Codeception/blob/4.1.6/src/Codeception/Module.php#L202-L266) and documented in https://codeception.com/docs/06-ModulesAndHelpers#Hooks
* It repeats methods of [DoctrineProvider](https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Lib/Interfaces/DoctrineProvider.php) and [PartedModule ](https://github.com/Codeception/Codeception/blob/4.1/src/Codeception/Lib/Interfaces/PartedModule.php) interfaces.
* `onReconfigure` method should be protected, it wasn't supposed to be called directly.
* Other methods are Symfony specific methods implemented by this module, so they are defined in Symfony.php file.

Naktibalda (changes requested):

documentation.md
```diff
@@ -235,6 +235,23 @@ $I->amOnPage('/register');
...
+]);
+$I->amLoggedInAs($user);
+```
+
+ * `param UserInterface` $user
+ * `param string` $firewallName
+ * `param null` $firewallContext
+
+
 ### amOnRoute
```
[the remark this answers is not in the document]
TavoNiievez: Fixed in [b3335e4](https://github.com/Codeception/module-symfony/pull/11/commits/b3335e4bada3b10a9311840001383eb831a34301). Thank you for explaining to me.

src/Codeception/Lib/Connector/Symfony.php
```diff
@@ -2,9 +2,9 @@
 namespace Codeception\Lib\Connector;
 
 use Symfony\Component\HttpKernel\HttpKernelBrowser;
+use Symfony\Component\HttpKernel\Kernel;
 
-//Alias for Symfony < 4.3
-if (!class_exists('Symfony\Component\HttpKernel\HttpKernelBrowser') && class_exists('Symfony\Component\HttpKernel\Client')) {
+if (Kernel::VERSION_ID < 40300) {
```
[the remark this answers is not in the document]
TavoNiievez: I think the same, sadly [it is not possible.](https://github.com/symfony/symfony/issues/35363)

src/Codeception/Module/Symfony.php
```diff
@@ -138,7 +138,8 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         'cache_router' => false,
         'em_service' => 'doctrine.orm.entity_manager',
         'rebootable_client' => true,
-        'mailer' => self::SWIFTMAILER
+        'mailer' => self::SWIFTMAILER,
+        'guard' => false
```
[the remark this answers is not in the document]
TavoNiievez: Fixed in [17f01e1](https://github.com/Codeception/module-symfony/pull/11/commits/17f01e1c294abbadb174a8c00a81cc4e24f2dd04).

src/Codeception/Module/Symfony.php
```diff
@@ -770,4 +771,31 @@ private function getPossibleKernelClasses()
 
         return [$this->config['kernel_class']];
     }
+
+    public function amLoggedInAs(UserInterface $user, $firewallName = 'main', $firewallContext = null)
```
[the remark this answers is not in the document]
TavoNiievez: Fixed in [7680646](https://github.com/Codeception/module-symfony/pull/11/commits/768064621b0929d6db77c909588cc0f443f00dfc).

src/Codeception/Module/SymfonyModuleInterface.php
TavoNiievez: Fixed in [34f06de](https://github.com/Codeception/module-symfony/pull/11/commits/34f06de2900e2714da8160ef5c6301e32cee396e). It is only for documentation, thanks to this interface it is clearer to see which public methods this module really provides and allows detecting details such as `onReconfigure` that had been overlooked, I hope it will be helpful.

TavoNiievez: @Naktibalda I was thinking, wouldn't it be great if the [Codeception page](https://codeception.com/docs/modules/Symfony) showed only the methods provided by this module (that are described in the interface)? In a future PR, would it be possible to move all that documentation from the code to the interface, and configure robo so that it only reads that documentation, without bringing the methods of the parent class?

DavertMik: Yeah, I agree with @Naktibalda that this interface is not needed unless there are different classes to implement.

As of documentation - if you make public method - it will be available in documentation. If you want it to be skipped just use underscore prefix _ to make this method hidden

Naktibalda: > I was thinking, wouldn't it be great if the Codeception page showed only the methods provided by this module (that are described in the interface)?

I think that it is a bad idea. The primary purpose of framework module is to implement connector for InnerBrowser which provides standard methods for functional testing.

Naktibalda (changes requested):

src/Codeception/Module/SymfonyModuleInterface.php
TavoNiievez: Fixed in [1e86a95.](https://github.com/Codeception/module-symfony/pull/11/commits/1e86a958b5e96d4b7e32778b5fc59d3766449e22) @DavertMik and @Naktibalda, yeap, i get your point, again, thanks for explaining 👍

composer.json
```diff
@@ -13,17 +13,17 @@
...
         "codeception/codeception": "^4.0"
     },
     "require-dev": {
         "codeception/util-robohelpers": "dev-master",
-        "codeception/module-asserts": "dev-master | ^1.0",
-        "codeception/module-doctrine2": "dev-master | ^1.0",
-        "codeception/module-phpbrowser": "dev-master | ^1.0",
-        "codeception/module-rest": "dev-master | ^1.0",
-        "codeception/module-sequence": "dev-master | ^1.0",
-        "vlucas/phpdotenv": "^3.6"
```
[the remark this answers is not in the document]
TavoNiievez: Fixed in [1fffcb1.](https://github.com/Codeception/module-symfony/pull/11/commits/1fffcb1a2d7013637c8800a2cb28c8101be361db)

src/Codeception/Module/Symfony.php
```diff
@@ -770,4 +774,47 @@ private function getPossibleKernelClasses()
...
+     * 'email' => 'john_doe@gmail.com'
+     * ]);
+     * $I->amLoggedInAs($user);
+     * ```
+     *
+     * @param UserInterface $user
+     * @param string $firewallName
+     * @param null $firewallContext
+     */
+    public function amLoggedInAs(UserInterface $user, $firewallName = 'main', $firewallContext = null)
```
c33s, line 795: as i wrote i the other PR, i am not sure if it is a good idea to go for the bleeding edge implementation of the login code.


since symfony 5.1 the login for testing is done like this:
```php
public function testVisitingWhileLoggedIn()
{
    $client = static::createClient();
    $userRepository = static::$container->get(UserRepository::class);

    // retrieve the test user
    $testUser = $userRepository->findOneByEmail('john.doe@example.com');

    // simulate $testUser being logged in
    $client->loginUser($testUser);

    // test e.g. the profile page
    $client->request('GET', '/profile');
    $this->assertResponseIsSuccessful();
    $this->assertSelectorTextContains('h1', 'Hello John!');
}
```
https://symfony.com/doc/5.1/testing/http_authentication.html
https://symfony.com/doc/5.1/testing.html#testing-logging-in-users


before you had to do this:
```php
public function loggedInAs($user = 'admin', $firewallContext = 'main', $roles = ['ROLE_SUPER_ADMIN', 'ROLE_ADMIN', 'ROLE_USER', 'ROLE_SONATA_ADMIN'])
{
    /** @var Session $session */
    $session = $this->getContainer()->get('session');
    $client = $this->getClient();

    $token = new UsernamePasswordToken($user, null, $firewallContext, $roles);
    $session->set('_security_'.$firewallContext, serialize($token));
    $session->save();
    $cookie = new Cookie($session->getName(), $session->getId());
    $client->getCookieJar()->set($cookie);
}
```
https://symfony.com/doc/5.0/testing/http_authentication.html

i have not tested if the new code is backward compatible but maybe there is a BC break. also i still think it is handy
to allow to login by role and not by user. both approaches are valid but in many of my projects i only care for roles a
user has and not for fetching a real user. i simply want to know if a admin can create/view/edit/delete an object while
an editor only is allowed to create/view and a user only is allowed to view.

so if the old approach also works for symfony 5.1+ i would go for the old approach and add some helper methods or a 2nd
method to allow to provide a user interface

composer.json
```diff
@@ -13,17 +13,17 @@
...
+        "codeception/module-asserts": "^1.3",
```
c33s, line 21: may i again ask you why you require higher versions for this? have you verified not to produce a BC break by increasing the versions?

composer.json
```diff
@@ -13,17 +13,17 @@
...
+        "codeception/module-doctrine2": "^1.0",
+        "codeception/module-phpbrowser": "^1.0",
+        "codeception/module-rest": "^1.2",
+        "codeception/module-sequence": "^1.0",
+        "vlucas/phpdotenv": "^4.1"
```
c33s: here you do a major version upgrade. why? is there a BC break? do you even touch the phpdotenv component in your PR?
if the new version is required can it coexist with the old version? `^3.6|^4.1`

composer.json
```diff
@@ -13,17 +13,17 @@
     "minimum-stability": "RC",
     "require": {
         "php": ">=5.6.0 <8.0",
-        "codeception/lib-innerbrowser": "^1.0",
+        "codeception/lib-innerbrowser": "^1.3",
         "codeception/codeception": "^4.0"
     },
     "require-dev": {
         "codeception/util-robohelpers": "dev-master",
-        "codeception/module-asserts": "dev-master | ^1.0",
```
c33s, line 21: i assume dev-master is also allowed if requested is it?

composer.json
```diff
@@ -2,7 +2,7 @@
     "name":"codeception/module-symfony",
     "description":"Codeception module for Symfony framework",
     "keywords":["codeception", "symfony"],
-    "homepage":"http://codeception.com/",
+    "homepage":"https://codeception.com/",
```
c33s, line 5: cosmetic changes still should be in their own PR

src/Codeception/Module/SymfonyModuleInterface.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
c33s: please drop the interface, it's simply not needed

src/Codeception/Module/Symfony.php
```diff
@@ -4,19 +4,19 @@
 
 use Codeception\Configuration;
 use Codeception\Exception\ModuleException;
-use Codeception\Lib\Framework;
```
c33s, line 7: such use reorderings are also easier to review if they happen in their own PR

src/Codeception/Module/Symfony.php
c33s: how to handle in memory user or basic auth? 
```yaml
security:
    providers:
        in_memory:
            memory:
                users:
                    mymemoryuser:
                        password: "my secret password"
                        roles: 'ROLE_SUPER_ADMIN'
```

having only the user interface option require me to mock a nonexisting user class or the interface.

c33s (changes requested):
as i am no member here i just post my personal opinion. this PR is much better than the last one but still has things touched which have nothing to do with the functionality. simply don't  do "and more" to the minimum and make the reviewers life easier. just my 2cts

src/Codeception/Module/Symfony.php
TavoNiievez: @c33s the same logic works in symfony 5.1. I try it. What the documentation says is that this logic is no longer necessary because the method you describe in symfony / phpunit-bridge was included by default. But that doesn't create any BC.

composer.json
TavoNiievez: this has already been requested in another comment, please do not spam.

src/Codeception/Module/SymfonyModuleInterface.php
TavoNiievez: again, this has already been commented, please do not spam.

src/Codeception/Module/Symfony.php
TavoNiievez: @c33s Regarding what you say about in_memory, in effect it is a totally valid and different form of authentication, and in fact it is already possible if you create a `security.yaml` file only for the `test` environment and get it to be imported in Codeception. I agree that it would be a good idea to add documentation on this.

TavoNiievez: @c33s Regarding what you say about in_memory, in effect it is a totally valid and different form of authentication, and it is possible if you get the `security.yml` file specific to the `test environment` to be imported or you modify the symfony configuration for the test on the way. I agree that it would be a good idea to add documentation on this.

TavoNiievez: @c33s Currently it is difficult to delegate tests to the CI server since the build fails in the installation of dependencies, this happens before this PR, and since before updating to Symfony 5.1 in [Codeception/symfony-demo](https://github.com/Codeception/symfony-demo), can you help to solve them? that way anyone who sends a PR could add the corresponding tests.

composer.json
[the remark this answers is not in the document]
TavoNiievez: Fixed in [e5c62ae.](https://github.com/Codeception/module-symfony/pull/11/commits/e5c62ae417e4694803a1e58b2bd726a280f2e568). Unfortunately version 5 is not supported yet by codeception and generates [the error mentioned in this issue.](https://github.com/Codeception/Codeception/issues/5946
)

composer.json
TavoNiievez: @c33s thanks you

src/Codeception/Module/Symfony.php
TavoNiievez: i had it in mind since then, thanks you

composer.json
TavoNiievez: @c33s not recommended

composer.json
TavoNiievez: @c33s yes, I have verified.

src/Codeception/Module/Symfony.php
TavoNiievez: @c33s Thanks for your interest in this, [thanks to this](https://github.com/Codeception/symfony-demo/commit/b469dcc81d65ef6c16fc5839bc1f07858730301d) everything works fine again, please let me know about any other concerns.
Regarding the function you are requesting, if you open an issue requesting that feature, I will gladly spend time developing another function with the specific API that you propose.
Cheers!

TavoNiievez:
@Naktibalda @DavertMik

DavertMik (approved):

DavertMik:
Looks good to me. I think we can merge it and publish next release!

TavoNiievez:
@Naktibalda

TavoNiievez:
@Naktibalda (?)

Naktibalda:
@TavoNiievez I merged 2 pull requests, not all others have merge conflicts.

TavoNiievez:
@Naktibalda I'll fix them today, thanks you

TavoNiievez:
@Naktibalda 
I lost the ability to update the fork i created these PRs with.
I'm sure the conflicts are white space issues at the end of the file,
but GitHub no longer gives me the permission to fix the conflicts :confused:

Naktibalda:
I merged these changes manually too.
I will make a few more changes and release a new version this weekend.

PR #12, opened 2020-09-10
Added logout function

ghost:
Just like the Laravel module, the Symfony Module should have this function. 

This time the implementation required something more elaborate than just calling `session->invalidate()` since although the session was removed the cookies were kept in `Symfony\Component\BrowserKit\AbstractBrowser` instance (`$this->client`).

It asks for the two cookies by default, but additionally it asks for the current session since they may not be the same.


TavoNiievez:
@c33s

PR #13, opened 2020-09-11
Added seeInSession function

ghost:
Just like in Laravel module, the Symfony Module should also have this function.

In this case the logic [is very similar](https://github.com/Codeception/module-laravel5/blob/bd2c604e8aa02d2b24737de01c6716473e8db96c/src/Codeception/Module/Laravel5.php#L610), the name of the parameters has been updated to be in accordance with the [Symfony documentation](https://symfony.com/doc/current/components/http_foundation/sessions.html).

As soon as this can be merged I will proceed to implement the `seeSessionHasValues` function to assert multiple attributes / values.

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -1092,6 +1092,21 @@ $I->seeInFormFields('//form[@id=my-form]', $form);
  * `param` $params
 
 
+### seeInSession
+ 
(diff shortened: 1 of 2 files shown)


TavoNiievez:
@c33s

PR #15, opened 2020-09-12
Added seeAuthentication function

ghost:
Added function that checks if a user is logged in, [based on Symfony documentation](https://symfony.com/doc/current/security.html#checking-to-see-if-a-user-is-logged-in-is-authenticated-fully) and corresponding roles.

The Laravel module [has a function](https://github.com/Codeception/module-laravel5/blob/bd2c604e8aa02d2b24737de01c6716473e8db96c/src/Codeception/Module/Laravel5.php#L792) with the same name, however the concept of `Guard` in Symfony has a very different implementation than Laravel.

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -866,6 +866,20 @@ For checking the raw source code, use `seeInSource()`.
  * `param array|string` $selector optional
 
 
+### seeAuthentication
+ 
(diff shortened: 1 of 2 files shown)


TavoNiievez:
@c33s

PR #21, opened 2020-10-22
Improved description of grabService

ThomasLandauer:
Explaining test.service_container

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -554,8 +554,9 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
-     * Grabs a service from Symfony DIC container.
-     * Recommended to use for unit testing.
+     * Grabs a service from the Symfony container.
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -554,8 +554,9 @@ public function dontSeeEmailIsSent()
     }
 
     /**
-     * Grabs a service from Symfony DIC container.
-     * Recommended to use for unit testing.
+     * Grabs a service from the Symfony container.
```
c33s, line 557: please leave DIC in the text

c33s (changes requested):

src/Codeception/Module/Symfony.php
ThomasLandauer: "dependency injection container"?

c33s: yes. sometimes you may search for DIC and then you will find this section :)

ThomasLandauer: Well, OK, but not even the Symfony Documentation finds anything when you search for "DIC" ;-) https://symfony.com/doc/current/index.html

PR #24, opened 2020-10-22
Docs: improved seeEmailIsSent description

ThomasLandauer:
Added info about Console, see https://github.com/Codeception/Codeception/issues/4698

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -475,9 +475,9 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     /**
      * Checks if the desired number of emails was sent.
      * If no argument is provided then at least one email must be sent to satisfy the check.
-     * The email is checked using Symfony's profiler. If your app performs a redirect after sending the email,
-     * you need to tell Codeception to not follow this redirect, using REST Module's [stopFollowingRedirects](
-     * https://codeception.com/docs/modules/REST#stopFollowingRedirects) method.
+     * The email is checked using Symfony's profiler, which means:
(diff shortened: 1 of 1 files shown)


TavoNiievez:
Hey @ThomasLandauer , you're right 😃, currently the way emails are verified is through the Profiler, and that requires a web request. So that's right, currently Codeception can't detect emails sent via CLI... But it could :). I can think of several ways to implement this:

- [Message Events.](https://github.com/symfony/symfony/pull/32927/files)
- [Checking email spool](https://stackoverflow.com/questions/32903655/test-email-sending-with-no-controller)
- [Persisting sent mails to a file.](https://stackoverflow.com/questions/37720439/symfony-testing-email-sending-in-command) Or...
- Through the 'logger' service as they do [here](https://stackoverflow.com/questions/31678775/get-content-of-email-sent-during-command-tests) for SwiftMailer or [here](https://github.com/symfony/framework-bundle/blob/e8656b179e41ee42010c150375d310d02af3a1af/Test/MailerAssertionsTrait.php#L121) in Symfony Mailer itself.

So this is another 'not yet implemented' issue, and i think it could be even better if an issue is opened requesting that feature. and instead of just saying "codeception can't do it", you could add a link to the issue in the description, so that people can see if someone has contributed that feature (and has been merged) or not.

ThomasLandauer:
Yeah, great! Judging from @DavertMik's answer (see above) I thought it's not possible in principle.

Would you create that issue? I can't say much - I would just copy your text from above... ;-)

TavoNiievez:
I personally would prefer to present the corresponding PR directly,
achieving this may involve editing various methods in the module, so i could try to implement it after the PRs i currently have open are merged or not (owners can ask me for changes there too).

ThomasLandauer:
Yeah, sure! I changed "Codeception cannot detect it." to "Codeception cannot detect it yet." - to give a hint that there's some progress underway.

So I'd say this docs PR here can be merged for now.

PR #29, opened 2020-11-07
Added seeUserPasswordDoesNotNeedRehash function

TavoNiievez:
Method to verify that the user's password would not benefit from rehashing.
```php
     $I->seeUserPasswordDoesNotNeedRehash();
     $I->seeUserPasswordDoesNotNeedRehash($user);
```
If the user is not specified it is taken from the session.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1074,6 +1074,44 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         $this->fail("Action '$action' does not exist");
     }
 
+    /**
+     * Checks that the user's password would not benefit from rehashing.
+     * If the user is not provided it is taken from the current session.
+     *
+     * You might use this function after performing tasks like registering a user or submitting a password update form.
+     *
+     * ```php
+     * <?php
+     * $I->seeUserPasswordDoesNotNeedRehash();
+     * $I->seeUserPasswordDoesNotNeedRehash($user);
+     * ```
+     *
+     * @param UserInterface|null $user
+     */
+    public function seeUserPasswordDoesNotNeedRehash(UserInterface $user = null)
+    {
+        $container = $this->_getContainer();
+
+        if (!$user) {
+            if (!$container->has('security.helper')) {
+                $this->fail("Symfony container doesn't have 'security.helper' service");
+            }
+
+            $security = $this->grabService('security.helper');
+            if (!$user = $security->getUser()) {
+                $this->fail('No user found to validate');
+            }
+        }
+
+        if (!$container->has('security.user_password_encoder.generic')) {
+            $this->fail("Symfony container doesn't have 'security.user_password_encoder.generic' service");
+        }
+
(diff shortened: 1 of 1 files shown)


DavertMik:
I don't understand the context fully. 
Could you specify where this method should be used?

DavertMik (approved):

TavoNiievez:
Sure. This method is useful after persisting entities that implement the `UserInterface`, for example when registering a new user, or updating their password,

With this function you can verify that the password is correctly encrypted with the best available algorithm. (that is, the listener or event that you wrote to encode the user's password is working correctly.)

src/Codeception/Module/Symfony.php
```diff
@@ -1074,6 +1074,56 @@ public function seeCurrentActionIs($action)
         $this->fail("Action '$action' does not exist");
     }
 
+    /**
+     * Checks that the user's password is encrypted
```
Naktibalda, line 1078: I find it odd that Symfony decided to name this feature password encoding
but using word "encrypted" is worse, because encrypting passwords is bad practice, they must be hashed.
I looked at Symfony Security component and it hashes passwords, except PlaintextPasswordEncoder.

src/Codeception/Module/Symfony.php
```diff
@@ -1074,6 +1074,56 @@ public function seeCurrentActionIs($action)
...
+        if ($rawPassword) {
+            $this->assertTrue($encoder->isPasswordValid($user, $rawPassword));
+            return;
+        }
+
+        if ($needsRehash = $encoder->needsRehash($user)) {
+            $this->markTestIncomplete('User password needs rehash');
+            return;
+        }
+        $this->assertFalse($needsRehash);
```
Naktibalda, line 1124: Your description says that this method `Checks that the user's password is encrypted`,
but I don't see that in the code.

If the second parameter is provided, this method checks if the password is correct.
otherwise it marks test as incomplete if password needs rehashing.

The behaviour of this method just doesn't match description,  
I would replace it with 2 separate methods - seeUserPasswordIsValid and skipTestIfPasswordNeedsRehash.

Naktibalda (changes requested):

src/Codeception/Module/Symfony.php
TavoNiievez: @Naktibalda Thank you!
Symfony does not provide a method to directly check if *a password **is** encoded*.

If for example, i have a password encoded with bcrypt, `needsRehash` would return `true`, because there are better algorithms like `argon2`... so i changed the name of the function.

Now there are **no** intermediate results with `markTestIncomplete` and i think that verifying that the password is encoded with the best possible algorithm is the only thing i want to assert.

src/Codeception/Module/Symfony.php
TavoNiievez: It's called 'encode' for historical reasons [[1](https://symfony.com/blog/new-in-symfony-4-3-native-password-encoder)], i agree that using 'encrypt' left even more room for improvement.

However in the new description there is no mention of that word, thank you for letting me know.

PR #31, opened 2020-11-07
Added checkSentEmail function

TavoNiievez:
Verifies that an $email instance is valid and was sent at least once.
```php
     $email = $this->mailer->sendEmail();
    
     $this->unitTester->checkSentEmail($email);
     $this->unitTester->checkSentEmail($email, [new Address(
         'john_doe@gmail.com', 'John Doe'
     )]);
```

**Edit**: I close this since I agree that it is easier to write it yourself.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1004,6 +1004,43 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Checks that an object is a valid \Email instance and was sent at least once.
+     *
+     * ```php
+     * <?php
+     *
+     * $email = $this->mailer->sendEmail();
+     *
+     * $this->unitTester->checkSentEmail($email);
+     * $this->unitTester->checkSentEmail($email, [new Address(
+     *     'john_doe@gmail.com', 'John Doe'
+     * )]);
+     * ```
+     *
+     * @param \Symfony\Component\Mime\Email $email
+     * @param \Symfony\Component\Mime\Address[]|null $addresses
+     * @param int $sent
+     */
+    public function checkSentEmail($email, $addresses = [], $sent = 1)
+    {
+        $emailClass = '\Symfony\Component\Mime\Email';
+        $addressClass = '\Symfony\Component\Mime\Address';
+        $recipients = $email->getTo();
+
+        $this->assertInstanceOf($emailClass, $email);
+
+        $this->assertGreaterOrEquals($sent, count($recipients));
+
+        $this->assertContainsOnlyInstancesOf($addressClass, $recipients);
+
+        if (!empty($addresses)) {
+            foreach ($recipients as $recipient) {
+                $this->assertContainsEquals($recipient, $addresses);
+            }
+        }
+    }
+
+    /**
      * Check that user is not authenticated.
      * You can specify whether users logged in with the 'remember me' option should be ignored by passing 'false' as a parameter.
      *


TavoNiievez:
@ThomasLandauer Can you please review this too? do you agree with the names?

ThomasLandauer:
Um, I fear I don't quite get what this is all about ;-) You pass an `$email` to this - but where do you get this from? And what means "valid" here - how could this be invalid? `$expected` is an array of just `Address`es or anything else? Is the function to be called before or after or instead of `seeEmailIsSent()`?

TavoNiievez:
@ThomasLandauer Thanks for asking :)

<details>
<summary>You pass an $email to this - but where do you get this from?</summary>
<br>

The `$email` object is in this case a Symfony Mailer Email object, [as you can see in the Symfony documentation](https://symfony.com/doc/current/mailer.html#creating-sending-messages), that is the object that you would normally send to the `$mailer->send();` method to send an email. 
I cannot type that argument (for compatibility reasons), but i did include the type in docblock.
```php
     * @param \Symfony\Component\Mime\Email $email
```
Would it be enough to add in the example the creation (or sending) of the `$email` object before calling `$I->seeEmailSentIsValid($email);` ?
</details>

<details>
<summary>And what means "valid" here - how could this be invalid?</summary>
<br>

By **'valid'** i mean that the following assertions are met:
- [x] The `$email` object that is received is an instance of the `Symfony\Component\Mime\Email` class, since you can also use it with subtypes like [TemplatedEmail](https://symfony.com/doc/current/mailer.html#html-content) or... your app may have an error and the object may be `null` when asserting. In that case the test would fail.
- [x] The mail was sent at least once.
- [x] The Addresses contained in the `$email->getTo()` method are objects of the `Symfony\Component\Mime\Address` class. See an example of how it's done in Symfony [here.](https://symfonycasts.com/screencast/mailer/unit-test#codeblock-5e9fb75299)
</details>

<details>
<summary>$expected is an array of just Addresses or anything else?</summary>
<br>

Just an optional array of `Address` objects.
```php
     * @param \Symfony\Component\Mime\Address[]|null $expected
```
</details>

<details>
<summary>Is the function to be called before or after or instead of seeEmailIsSent()?</summary>
<br>

After.
I can think of improving the description of the `$I->seeEmailSentIsValid();` function to include the time in which it should be called, do you agree?
</details>

ThomasLandauer:
So this is for *unit* tests, not functional tests!? Cause the `$email` is created by the app, not in the test?!
The reason for me being so "slow" here is that my own unit tests mainly contain stupid, err..., "low-level" assertions like `assertSame()`, and I wouldn't expect Codeception to provide convenience functions like this. Are there similar functions like this integrated yet? If yes, there should be a header "Methods for Unit Tests" (or so) at https://codeception.com/docs/modules/Symfony. If this is (and will be) the only one, we can maybe just mention it in the description.

Sorry, but before I can answer to your answers :-) I need to get the big picture...

TavoNiievez:
@ThomasLandauer Thanks for your feedback, I hope the changes bring more clarity than before.

ThomasLandauer:
OK, but in a unit test, you can't really verify that the email actually has been sent - `$this->assertGreaterOrEquals($sent, count($recipients));` https://github.com/Codeception/module-symfony/pull/31/files#diff-7fbe8f18ab69169d0c4e82c0d30318f7b42b6620d49c59c7db8176a9cbe08161R1032 only checks if an address is present - not if the email was successfully handed over to some transport - right?

Then a better name would be `seeEmailIsValid()`. But: An email can be *invalid* in a thousand different ways - most of which Symfony will take care of. At least I hope :-) So I think it's hard to explain that this function focuses on two aspects (`instanceof Email` and the addresses), and not on some others (e.g. `Date` header, `Message-ID` header, attachments, total size ...)

So I'm not convinced that the effort it takes for possible users to understand what this function really does, is less than writing the assertions by themselves.

So before putting more effort into it, I'd suggest to wait for feedback from others :-)

PR #32, opened 2020-11-07
Added submitSymfonyForm function

TavoNiievez:
While testing forms in Symfony it is very common to have to repeat the form name in different places in the method to comply with **the Symfony naming format**:
```php
	$I->submitForm('form[name={form_name}]', [
		'{form_name}[{field_key}]' => '{field_value}',
		'{form_name}[{field_key}]' => '{field_value}'
	], '{form_name}_submit');
```
However this **ruins the readability of the test** and makes the code much more verbose:
```php
	$I->submitForm('form[name=user_update_password_form]', [
		'user_update_password_form[oldPassword]'      => '123456',
		'user_update_password_form[password][first]'  => '654321',
		'user_update_password_form[password][second]' => '654321'
	], 'user_update_password_form_submit');
```
That is why i write this function that allows you to take advantage of the fact that the naming format **is always the same**, and allows you to *simplify* the action of submitting a form:
```php
	$I->submitSymfonyForm('user_update_password_form', [
		'[oldPassword]'      => '123456',
		'[password][first]'  => '654321',
		'[password][second]' => '654321'
	]);
```
For the few who **do** edit this format in their applications and create something super-custom there is always the option to use the base `->submitForm()` function, so this is basically a shortcut alias function.

The improvement in readability is significant.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -966,6 +966,35 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Submits the given Symfony Form on the page, with the given form values.
+     * If you customized the CSS selectors or field names use ->submitForm() instead.
+     *
+     * ```php
+     * <?php
+     * $I->submitSymfonyForm('login_form', [
+     *     '[email]'    => 'john_doe@gmail.com',
+     *     '[password]' => 'secretForest'
+     * ]);
+     * ```
+     *
+     * @param string $name
+     * @param string[] $fields
+     */
+    public function submitSymfonyForm($name, $fields)
+    {
+        $selector = sprintf('form[name=%s]', $name);
+
+        $params = [];
+        foreach ($fields as $key => $value) {
+            $fixedKey = sprintf('%s%s', $name, $key);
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -965,6 +965,35 @@ public function seeAuthentication($remembered = false)
         $this->assertTrue($security->isGranted($role), 'There is no authenticated user');
     }
 
+    /**
+     * Submits the given Symfony Form on the page, with the given form values.
```
DavertMik, line 969: from this description it is not clear when should I use submitSimfonyForm and submitForm
so a small description why you would need to use this method over `submitForm` is required

DavertMik (changes requested):

src/Codeception/Module/Symfony.php
TavoNiievez: How about this:

> **Submit a form specifying the form name only once.**
> 
> **Use this function instead of `$I->submitForm()` to avoid repeating the form name in the field selectors.**
> 
> **If you customized the names of the field selectors use `$I->submitForm()` for full control.**

PR #33, opened 2020-11-08
Added availability check functions

TavoNiievez:
`$I->seePageIsAvailable()` allows to do in a line of code what is normally done like this: [[See source](https://github.com/symfony/demo/blob/bb3b1834079e1af8a65786b48415c8147fc7ee10/tests/Controller/DefaultControllerTest.php#L52)]
```php
    public function testPublicBlogPost(): void
    {
        $client = static::createClient();
        $blogPost = $client->getContainer()->get('doctrine')->getRepository(Post::class)->find(1);
        $client->request('GET', sprintf('/en/blog/posts/%s', $blogPost->getSlug()));

        $this->assertResponseIsSuccessful();
    }
```

while `$I->seePageRedirectsTo()` allows to do in a line of code what is normally done in this way: [[See source](https://github.com/symfony/demo/blob/bb3b1834079e1af8a65786b48415c8147fc7ee10/tests/Controller/DefaultControllerTest.php#L69)]

```php
    public function testSecureUrls(string $url): void
    {
        $client = static::createClient();
        $client->request('GET', $url);

        $this->assertResponseRedirects(
            'http://localhost/en/login',
            Response::HTTP_FOUND,
            sprintf('The %s secure URL redirects to the login form.', $url)
        );
    }
```

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -476,6 +476,45 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Goes to a page and check that it can be accessed.
+     *
+     * ```php
+     * <?php
+     * $I->seePageIsAvailable('/dashboard');
+     * ```
+     *
+     * @param string $url
+     */
+    public function seePageIsAvailable($url)
+    {
+        $this->amOnPage($url);
+        $this->seeResponseCodeIsSuccessful();
+        $this->seeInCurrentUrl($url);
+    }
+
+    /**
+     * Goes to a page and check that it redirects to another.
+     *
+     * ```php
+     * <?php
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -475,6 +475,45 @@ public function seeInCurrentRoute($routeName)
...
+     * Goes to a page and check that it can be accessed.
+     *
+     * ```php
+     * <?php
+     * $I->seePageIsAvailable('/dashboard');
+     * ```
+     *
+     * @param string $url
+     */
+    public function seePageIsAvailable($url)
```
DavertMik, line 488: Maybe we can take this into InnerBrowser so it would be available to all frameworks and php browser? 
This doesn't seem to be Symfony related

DavertMik (approved):

PR #40, opened 2020-11-16
Update Symfony.php

ThomasLandauer:
Adding link to central "Parts" chapter, as announced in https://github.com/Codeception/module-webdriver/pull/33#issue-519184018
If you agree, I'll copy-paste this to all other modules that have parts.


TavoNiievez:
much better now, thanks @ThomasLandauer

PR #45, opened 2020-11-19
Added seeInFlashBag function

TavoNiievez:
This function verifies that there is a flash message in the session. See docblock for full description.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1260,4 +1260,33 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
 
         return $em->getRepository($mixed);
     }
+
+    /**
+     * Verifies that there is a flash message in the session.
+     * * If your app performs a redirect after setting the flash message, you need to suppress this using REST Module's [stopFollowingRedirects](https://codeception.com/docs/modules/REST#stopFollowingRedirects)
+     *
+     * ```php
+     * <?php
+     * $I->seeInFlashBag('success');
+     * $I->seeInFlashBag('success', '¡User successfully registered!');
+     * ```
+     *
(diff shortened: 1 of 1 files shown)


Naktibalda:
Is this method really necessary?

Flash message should be rendered by the next page and it can be tested using `see` method, right?
There is no need to stop redirects too .

TavoNiievez:
@Naktibalda , thanks for asking.
Please refer to this issue. https://github.com/Codeception/Codeception/issues/5351
There are difficulties asserting flash messages in functional tests due to limitations of phpbrowser.

Although this method does not verify that the message is displayed on the screen as it would with `see`, it does verify that the algorithm flow and the symfony session/flashbag are working correctly and as expected. 
I think that is what you would like to be able to verify in a functional/framework test.
While the message that is shown on the screen and with the expected selector is what should be verified in an acceptance/browser test.

TavoNiievez:
Regarding the need to stop redirects... Yes, it is necessary.
Symfony flash bag [works through stacks](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/HttpFoundation/Session/Flash/FlashBagInterface.php), which means that as soon as a message is accessed (in this case, displayed in a view) it is removed from the stack. So if you do a redirect the message will not exist in the flashbag at the time of asserting.

Naktibalda:
#5351 requires further investigation, not workaround.

TavoNiievez:
I agree

TavoNiievez:
@Naktibalda It seems that my "pandemic self" had a hunch that this was a real shortcoming. Symfony added a method for this:

https://symfony.com/blog/new-in-symfony-8-1-dx-improvements-part-1#asserting-flash-messages-in-tests

so, since this module inherits the assertions from Symfony, this will be implemented anyway.

PR #48, opened 2020-11-20
Added seeFormHasErrors function

TavoNiievez:
It is [also present](https://github.com/Codeception/module-laravel5/blob/1d8a82f78a6e8c26f49af65d9001fa311785d54b/src/Codeception/Module/Laravel5.php#L661) in the Laravel Module,

Check for errors after submitting a form.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1171,6 +1171,27 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Verifies that there are one or more errors bound to the submitted form.
+     *
+     * ``` php
+     * <?php
+     * $I->seeFormHasErrors();
+     * ```
+     *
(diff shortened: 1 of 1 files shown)


Naktibalda:
I can't find any documentation of form collector. Is it documented at all?

TavoNiievez:
It has no page on symfony.com and it is referenced very little (even on stackoverflow), that made it difficult to write these functions... 
```php
    /**
     * Returns all collected data.
     *
     * @return array|Data
     */
    public function getData();
```
But i debugged the 'Data' object deeply until i found the info i needed. there are no other objects in it, only arrays.

TavoNiievez:
I verified that those keys were the correct ones, doing a bit of reverse engineering [here](https://github.com/symfony/web-profiler-bundle/blob/5.x/Resources/views/Collector/form.html.twig).

Naktibalda (approved):

PR #50, opened 2020-11-20
Added seeFormErrorMessage function

TavoNiievez:
```php
     $I->seeFormErrorMessage('username');
     $I->seeFormErrorMessage('username', 'Username is empty');
```
[Also present](https://github.com/Codeception/module-laravel5/blob/1d8a82f78a6e8c26f49af65d9001fa311785d54b/src/Codeception/Module/Laravel5.php#L726) in the Laravel module... But a bit more complex 😃.

- Fails if there are no forms when asserting.
- Fails if an invalid field is passed as an argument.
- It performs its function by verifying that the field is valid and has errors.
- If the expected error message is specified it verifies that it matches the current one. (technically with: `->assertStringContainsString()`).

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1122,6 +1122,65 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Verifies that a form field has an error.
+     * You can specify the expected error message as second parameter.
+     *
+     * ``` php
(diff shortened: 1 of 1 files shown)


Naktibalda (approved):

DavertMik:
Good. Once it is merged we should have corresponding tests in symfony-demo

DavertMik (approved):

PR #51, opened 2020-11-22
Remove outdated php5 logic

TavoNiievez:
Now that the minimum version of php for this module is php 7.0.8 this conditional is never used


Naktibalda (approved):
Looks good to me

PR #53, opened 2020-11-23
Test the module in symfony 3.4 and php7.3+

TavoNiievez:
This module is currently not being tested in Symfony 3.4 on PHP 7.3 and PHP 7.4 (and soon neither on php8) due to a error (you can check it in the CI Build).

Apparently this error originates from the test project code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,10 +25,6 @@ jobs:
         php: [7.0, 7.1, 7.2, 7.3, 7.4]
         symfony: [3.4, 4, 5]
         exclude:
-          - php: 7.3
-            symfony: 3.4
-          - php: 7.4
-            symfony: 3.4
           - php: 7.0
             symfony: 4
           - php: 7.0


Naktibalda:
That error happens because that branch of symfony-demo uses version of Doctrine incompatible with PHP 7.3+

.travis.yml
```diff
@@ -16,10 +16,6 @@ env:
     - VERSION=5 TEST_REPO='-b symfony5 https://github.com/Codeception/symfony-demo.git' SUITES=functional,unit
 matrix:
   exclude:
-    - php: 7.3
-      env: VERSION=3.4 TEST_REPO='--recurse-submodules https://github.com/Naktibalda/codeception-symfony-tests'
-    - php: 7.4
```
Naktibalda, line 21: @eXorus deleted .travis.yml in other modules during migration to Github Actions.
I think that this file survived by accident and can be deleted.

eXorus: Yes I confirmed.

TavoNiievez: I have removed Travis.
I am also closing this PR due to work started on #54, thanks for your feedback.

PR #56, opened 2020-11-24
Fix seeEmailIsSent 

### LINKED ISSUE #55: Bug with 1.3 seeEmailIsSent() : Failed asserting that 0 is greater than 0.
Hi,

When updated the symfony module to 1.3, I have errors when I use function : **seeEmailIsSent()**

I can't use 0 in expected count : **seeEmailIsSent(0)** or use **dontSeeEmailIsSent()**, both throw an error : Failed asserting that 0 is greater than 0.

TavoNiievez:
Closes #55

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -548,7 +548,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
             $realCount = count($mailCollector->getEvents()->getMessages());
         }
 
-        if ($expectedCount) {
+        if ($expectedCount !== null) {
             $this->assertEquals($expectedCount, $realCount, sprintf(
                 'Expected number of sent emails was %d, but in reality %d %s sent.',
                 $expectedCount, $realCount, $realCount === 2 ? 'was' : 'were'


src/Codeception/Module/Symfony.php
```diff
@@ -548,7 +548,7 @@ public function seeEmailIsSent($expectedCount = null)
             $realCount = count($mailCollector->getEvents()->getMessages());
         }
 
-        if ($expectedCount) {
+        if ($expectedCount !== null) {
             $this->assertEquals($expectedCount, $realCount, sprintf(
                 'Expected number of sent emails was %d, but in reality %d %s sent.',
                 $expectedCount, $realCount, $realCount === 2 ? 'was' : 'were'
```
Naktibalda, line 554: Shouldn't this condition be `$realCount === 1 ? 'was' : 'were'` ?

TavoNiievez: @Naktibalda no, when there are two emails in the email collector, only one was actually sent.

TavoNiievez: @Naktibalda Hmm, apparently it's been like this since the function [was created.](https://github.com/Codeception/module-symfony/commit/9c52178431b6e743538c18202702b35d13056f68) I'm not sure if it has to do with how Swift Mailer vs Symfony Mailer works.

should i change it?

PR #61, opened 2020-11-26
Support PHP 8

Naktibalda:
All dependencies actually work on PHP8, some of them just haven't updated version requirement in their composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "minimum-stability": "RC",
     "require": {
-        "php": ">=7.0.8 <8.0",
+        "php": ">=7.0.8 <9.0",
         "codeception/lib-innerbrowser": "^1.3",
         "codeception/codeception": "^4.0",
         "ext-json": "*"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -12,7 +12,7 @@
     ],
     "minimum-stability": "RC",
     "require": {
-        "php": ">=7.0.8 <8.0",
+        "php": ">=7.0.8 <9.0",
```
TavoNiievez, line 15: Compatibility with php 8.0 does not guarantee compatibility with 8.1+ .
I personally would have blocked it at `<8.1`. But I've seen that that was the way to add PHP8 support elsewhere in Codeception.

TavoNiievez (approved):

composer.json
Naktibalda: I'm not doing this every year 😄 
There were no compatibility issues with 8.0 in module code across all modules, so I have high expectations for 8.1 too.

PR #64, opened 2020-12-02
Adding `dontSeeInSession()`

ThomasLandauer:
`seeInSession()` was already there, but the opposite was missing.

I omitted the `$value` parameter for this reason: If `$attribute` is *not* in the session, it doesn't make much sense to do an additional check about its (inexistent) value. Or does it?

Tests are missing, but I don't know how to do that :-(

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -935,7 +935,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
 
         if (null === $value) {
             if ($session->has($attribute)) {
-                $this->fail("No session attribute with name '$attribute'");
+                $this->fail("Session attribute with name '$attribute' does exist");
             }
         }
         else {
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -917,6 +917,26 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Assert that a session attribute does not exist.
+     *
+     * ```php
+     * <?php
+     * $I->dontSeeInSession('attribute');
+     * ```
+     *
+     * @param string $attribute
+     * @return void
+     */
+    public function dontSeeInSession(string $attribute)
+    {
+        $session = $this->grabService('session');
+
+        if ($session->has($attribute)) {
+            $this->fail("Session attribute with name '$attribute' does exist");
+        }
+    }
+
+    /**
      * Opens web page by action name
      *
      * ``` php
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -917,22 +917,29 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
-     * Assert that a session attribute does not exist.
+     * Assert that a session attribute does not exist, or is not equal to the passed value.
      *
      * ```php
      * <?php
      * $I->dontSeeInSession('attribute');
+     * $I->dontSeeInSession('attribute', 'value');
      * ```
      *
      * @param string $attribute
+     * @param mixed|null $value
      * @return void
      */
-    public function dontSeeInSession(string $attribute)
+    public function dontSeeInSession(string $attribute, $value = null)
     {
         $session = $this->grabService('session');
 
-        if ($session->has($attribute)) {
-            $this->fail("Session attribute with name '$attribute' does exist");
+        if (null === $value) {
+            if ($session->has($attribute)) {
+                $this->fail("No session attribute with name '$attribute'");
+            }
+        }
+        else {
+            $this->assertNotEquals($value, $session->get($attribute));
         }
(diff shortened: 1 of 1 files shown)


TavoNiievez:
Hey @ThomasLandauer

Thanks!

> it doesn't make much sense to do an additional check about its (inexistent) value. Or does it?

<details>
<summary>Clic to see answer.</summary>
<br/>

Regarding this, **i do** think that it would make sense.

This function actually does two things in its affirmative version: check attributes and check values.

Perhaps what i want to assert is that there is **no** specific **value** in a session attribute. What is the same, it does not matter to me if the attribute exists or not, **but if it does exist** i want the test to fail if it contains the value that i specify as the second parameter in 'dontSeeInSession'.
</details>

> Tests are missing, but I don't know how to do that :-(

<details>
<summary>Clic to see answer.</summary>
<br/>

This should be fixed.

My idea when re-writing the module's test project and adding tests is that from the moment that that can be released, anyone who wants to include new functions (including me) should also contribute the corresponding tests.
So personally i find it convenient to create an CONTRIBUTING.md file with the contribution guidelines and an explanation of how to test the code in the module's test project.

Following that order of ideas, the process for adding tests would include fork the test project, patching the composer.json so that your module fork is downloaded instead of the official version of the module, and modifying the SymfonyModuleCest.php file by adding the corresponding test.
Is there anything else you would like to see in that contribution file if it is created?
</details>

ThomasLandauer:
1. `$value`: If I wanted to assert that, I'd grab the value and use `assertNotEquals()`. But there is no `grabFromSession()`! So the more important question I'm seeing is: Should there be a `grabFromSession()`? Or should we tell people to just use `grabService('session');` and go with that? Which leads to the more general question: If the framework offers a convenient way for something, should Codeception duplicate it with its own function? I raised this question before about `grabEntityFromRepository()` at https://github.com/Codeception/Codeception/issues/5439#issuecomment-557905843

2. `CONTRIBUTING.md`: I haven't found a decent workflow for GitHub PR's for myself yet, that's why I'm always doing it on the website ;-) So if you could come up with a step-by-step instruction, I'd be happy to be its first user :-)

TavoNiievez:
@ThomasLandauer 

1. As i see it, - getting the service, - then getting the attribute and, - finally asserting you do 3 instructions, while with seeInSession / dontSeeInSession you do it all in one line, it's just syntax sugar, the point of having these functions is that anyone can read the test without having to know the implementation details of the framework.
Following your logic, there would not be many 'really necessary' methods in this module. Because most just get a service, do (or don't) do a couple of validations, and assert. Anyone can write that. But Codeception makes it easier to read and write.
Specifically, in this case, seeInSession exists because there are other modules that have it, and Codeception is supposed to be a framework agnostic tool. So...

> Should there be a grabFromSession()?

Nop.

2. Of course! I'll add it to my to-do list :-)

src/Codeception/Module/Symfony.php
```diff
@@ -916,6 +916,33 @@ public function seeInSession(string $attrib, $value = null)
...
+     * @param mixed|null $value
+     * @return void
+     */
+    public function dontSeeInSession(string $attribute, $value = null)
+    {
+        $session = $this->grabService('session');
+
+        if (null === $value) {
+            if ($session->has($attribute)) {
+                $this->fail("No session attribute with name '$attribute'");
```
TavoNiievez, line 938: This error message needs to be updated.
Remember that this means that the attribute value was found in the session.

TavoNiievez (changes requested):

src/Codeception/Module/Symfony.php
TavoNiievez: Just one more thing.
This error message needs to be updated.
Remember that this conditional indicates that the attribute was found in the session and this function verifies the opposite.

TavoNiievez, line 938: Just one more thing.
This error message needs to be updated.
Remember that this conditional indicates that the attribute was found in the session and this function verifies the opposite.

TavoNiievez (changes requested):

ThomasLandauer:
Well, you (sort of) convinced me in this case - I added the `$value` ;-)

However, I just suggested to deprecate `grabEntityFromRepository()`: https://github.com/Codeception/module-doctrine2/issues/22

src/Codeception/Module/Symfony.php
ThomasLandauer: Sorry for copy-pasting too eagerly :-)

TavoNiievez (approved):

ThomasLandauer (commented):
I clicked "Start a review" - which probably wasn't the right thing to do in this case...

ThomasLandauer:
Didn't you leave a *second* line comment?? It's gone now... :-(

TavoNiievez:
@ThomasLandauer don't worry and thanks you.
I tested it locally and it works quite well. 
I will write the test when this gets merged and while i work on creating the contribution guidelines.

PR #67, opened 2020-12-07
Added seeFormErrorMessages function

TavoNiievez:
It is [also present](https://github.com/Codeception/module-laravel5/blob/1d8a82f78a6e8c26f49af65d9001fa311785d54b/src/Codeception/Module/Laravel5.php#L701) in the Laravel module.

A validation was added to allow omitting `=> null` in the array if you don't want to specify the error message.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1347,6 +1347,32 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Verifies that a form has fields with errors.
+     *
+     * This method calls `seeFormErrorMessage` for each entry in the `$bindings` array.
+     *
+     * ``` php
+     * <?php
+     * $I->seeFormErrorMessages(['telephone', 'address']);
+     * $I->seeFormErrorMessages([
+     *     'telephone' => 'Phone number too short',
+     *     'address'   => null
+     * ]);
+     * ```
+     * @param string[] $bindings
+     */
+    public function seeFormErrorMessages(array $bindings)
+    {
+        foreach ($bindings as $key => $value) {
+            if (is_int($key)) {
+                $this->seeFormErrorMessage($value);
+            } else {
+                $this->seeFormErrorMessage($key, $value);
+            }
+        }
+    }
+
+    /**
      * Verifies that a form field has an error.
      * You can specify the expected error message as second parameter.
      *
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1347,27 +1347,55 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
-     * Verifies that a form has fields with errors.
+     * Verifies that multiple fields on a form have errors.
      *
-     * This method calls `seeFormErrorMessage` for each entry in the `$bindings` array.
+     * If you only specify the name of the fields, this method will
+     * verify that the field contains at least one error of any type:
      *
      * ``` php
      * <?php
      * $I->seeFormErrorMessages(['telephone', 'address']);
+     * ```
+     *
+     * If you want to specify the error messages, you can do so
+     * by sending an associative array instead, with the key being
+     * the name of the field and the error message the value.
+     *
+     * This method will validate that the expected error message
+     * is contained in the actual error message, that is,
+     * you can specify either the entire error message or just a part of it:
+     *
+     * ``` php
+     * <?php
      * $I->seeFormErrorMessages([
-     *     'telephone' => 'Phone number too short',
-     *     'address'   => null
+     *     'address'   => 'The address is too long'
+     *     'telephone' => 'too short', // the full error message is 'The telephone is too short'
      * ]);
      * ```
-     * @param string[] $bindings
+     *
+     * If you don't want to specify the error message for some fields,
+     * you can pass `null` as value instead of the message string,
+     * or you can directly omit the value of that field. If that is the case,
+     * it will be validated that that field has at least one error of any type:
+     *
+     * ``` php
+     * <?php
+     * $I->seeFormErrorMessages([
+     *     'telephone' => 'too short',
+     *     'address'   => null,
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -1108,6 +1108,32 @@ public function seeCurrentActionIs(string $action)
...
+     * @param string[] $bindings
+     */
+    public function seeFormErrorMessages(array $bindings)
+    {
+        foreach ($bindings as $key => $value) {
+            if (is_int($key)) {
+                $this->seeFormErrorMessage($value);
+                continue;
+            }
+            $this->seeFormErrorMessage($key, $value);
```
Naktibalda, line 1133: Style comment.
I think that in this situation `else` would look much better and would make the code easier to understand.

```php
            if (is_int($key)) {
                $this->seeFormErrorMessage($value);
            } else {
                $this->seeFormErrorMessage($key, $value);
            }
```

TavoNiievez: I'm not used to writing `else` statements,
but i changed it as you proposed.
thanks

src/Codeception/Module/Symfony.php
```diff
@@ -1108,6 +1108,32 @@ public function seeCurrentActionIs(string $action)
...
+     * <?php
+     * $I->seeFormErrorMessages(['telephone', 'address']);
+     * $I->seeFormErrorMessages([
+     *     'telephone' => 'Phone number too short',
+     *     'address'   => null
+     * ]);
+     * ```
+     * @param string[] $bindings
+     */
+    public function seeFormErrorMessages(array $bindings)
```
Naktibalda, line 1126: I don't like this method at all.

It is unclear what this method does without reading description of `seeFormErrorMessage`.

$bindings can be used in 2 different formats - in the first format field names are passed as array values, in the second format field names are passed as keys.  
Behaviour of different formats is not documented, I doubt that anyone will understand it without reading the source code.

I see some benefits in `$I->seeFormErrorMessages(['telephone', 'address']);` - it can be used to check that multiple fields have errors in one line of code.

But second format makes no sense to me.
It adds 2 extra lines of code to just calling `seeFormErrorMessage` directly:

```php
$I->seeFormErrorMessage('telephone', 'Phone number too short');
$I->seeFormErrorMessage('address');
```

Variable names are meaningless too, but I will leave that for another time.

TavoNiievez: **Click to see the answers in each case.**

<details>
<summary>I don't like this method at all.</summary>
<br/>

Ok, what i understand you mean is that the changes you are asking for are:
- Add a text to the docblock that explains what `seeFormErrorMessage` does.
- Add more examples on the use of the function, especially in its expanded form.
<hr/>
</details>

<details>
<summary>But second format makes no sense to me.</summary>
<br/>

Regarding this, the utility that i see is not to repeat many `seeFormErrorMessage`.
Imagine that i have a form with 9 fields, and for each of them i have 7 validations. (too short, too long, cannot contain numbers, etc ...) There are a total of **63 rules** to validate for a single form...
So... Instead of writing 63 times:
```php
$I->seeFormErrorMessage('telephone', 'Phone number too short');
$I->seeFormErrorMessage('telephone', 'Phone cannot contain letters');
$I->seeFormErrorMessage('address', 'Address is too long');
// 60 more calls to $I->seeFormErrorMessage
```
You could write `seeFormErrorMessages`. The gain here is in legility:
```php
$I->seeFormErrorMessages([
	'telephone' => 'Phone number too short',
	'telephone' => 'Phone cannot contain letters',
	'address'   => 'Address is too long',
	// more fields in single call
);
```
With the short form it is true that i can assert that there is an error with that field, but **a field can have multiple validations**, it may be showing an error for another reason or for **several** and that is what i would like to know in my tests.
<hr/>
</details>


<details>
<summary>Variable names are meaningless too, but I will leave that for another time.</summary>
<br/>

I agree that there is room for improvement for the name of the variables, but please, do not leave it for later. If you have an idea of ​​which name would be better for that parameter please tell me to make the change.

I took into account how it was done in the Laravel module and decided to keep that name because i couldn't think of a better one.
<hr/>
</details>

<details>
<summary>I doubt that anyone will understand it without reading the source code.
</summary>
<br/>

All the observations you made apply to the method in the Laravel Module as well, and changes made here could benefit both modules. 

However, i think **i need some extra help here.** I wouldn't know exactly how to write a docblock that meets your requirements without repeating text, adding too much text, adding a lot of examples, or using the proper names. I could change it a few times until you tell me it's okay but that doesn't seem like a very efficient way of doing things.

For the moment i'll add a 'help wanted' label here, and i'll create an issue in the Laravel module where this method is already added with all these flaws, so we don't lose sight of it.
<hr/>
</details>

Naktibalda:
```php
$I->seeFormErrorMessages([
	'telephone' => 'Phone number too short',
	'telephone' => 'Phone cannot contain letters',
	'address'   => 'Address is too long',
	// more fields in single call
);
```
You can't have two items with key `telephone` in array.

TavoNiievez: ups 😆 
that's correct. It still reads better with unique fields in a single call.

DavertMik:
Thanks @TavoNiievez for PR
Thanks @Naktibalda for a review

My biggest concern here that this method validates full error messages, which is useless in most of cases. Those error messages are either standard (so no reason to test the full message). 

Consider the case - we try to improve error messages, and we change them on each iteration adding more and more descriptinve erros and more and more polite languages, while keeping validation rules the same. Should we update test on each wording change? I prefer not. So personally I would assert for the error field to *include* some text.

However, not sure how to do this correctly with consistent API

TavoNiievez:
@DavertMik if i understand you correctly, what you are asking is that an `assertStringContainsString` be made to be able to pass only a fragment of the error message instead of the whole message. That sounds better to me too.

TavoNiievez:
@Naktibalda @DavertMik I've made all the requested changes based on your feedback,

thanks to both of you.

PR #68, opened 2020-12-09
Added seeEventTriggered function

TavoNiievez:
It [is also present](https://github.com/Codeception/module-laravel5/blob/1d8a82f78a6e8c26f49af65d9001fa311785d54b/src/Codeception/Module/Laravel5.php#L374) in the Laravel module.

The symfony translation was a bit more complex, since there is no method to directly ask if an event was fired.

In any case, i was able to get the same behavior.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -1079,6 +1079,43 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     }
 
     /**
+     * Make sure events fired during the test.
+     *
+     * ``` php
+     * <?php
+     * $I->seeEventTriggered('App\MyEvent');
+     * $I->seeEventTriggered(new App\Events\MyEvent());
+     * $I->seeEventTriggered(['App\MyEvent', 'App\MyOtherEvent']);
+     * ```
+     * @param string|object|string[] $expected
+     */
+    public function seeEventTriggered($expected)
+    {
+        $eventCollector = $this->grabCollector('events', __FUNCTION__);
+
+        $data = $eventCollector->getCalledListeners();
+
+        if ($data->count() === 0) {
+            $this->fail('No event was triggered');
+        }
+
+        $actual = $data->getValue(true);
+        $expected = is_array($expected) ? $expected : [$expected];
+
+        foreach ($expected as $expectedEvent) {
+            $triggered = false;
+            $expectedEvent = is_object($expectedEvent) ? get_class($expectedEvent) : $expectedEvent;
+
+            foreach ($actual as $actualEvent) {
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -1078,6 +1078,44 @@ public function grabParameter(string $name)
...
+     * ``` php
+     * <?php
+     * $I->seeEventTriggered('App\MyEvent');
+     * $I->seeEventTriggered(new App\Events\MyEvent());
+     * $I->seeEventTriggered(['App\MyEvent', 'App\MyOtherEvent']);
+     * ```
+     * @param string|object|string[] $expected
+     * @param bool $orphan
+     */
+    public function seeEventTriggered($expected, bool $orphan = false)
```
Naktibalda, line 1093: What is the `$orphan` parameter for? It isn't documented.  
Using boolean flag parameters is considered to be a bad practice,
it would be better to introduce another method, e.g. `seeOrphanedEventTriggered`.

TavoNiievez: Ok.
Before making the changes and creating the new function, i would like to clarify something:

> Using boolean flag parameters is considered to be a bad practice

To be consistent with this, do you agree to create specific functions for the following methods that already include parameter flags?:

```php
public function persistService(string $serviceName, bool $isPermanent = false)
public function seeAuthentication(bool $remembered = false)
public function dontSeeAuthentication(bool $remembered = true)
```
Something like:
`persistPermanentService`
`seeRememberedAuthentication`
`dontSeeRememberedAuthentication`

If we know that it is a bad practice, it is something that we must correct, right?

TavoNiievez: One more thing.
Should the name be `seeOrphanEventTriggered` or `seeOrphanedEventTriggered` ?

TavoNiievez: @Naktibalda One more thing.
Should the name be `seeOrphanEventTriggered` or `seeOrphanedEventTriggered` ?

Naktibalda: > To be consistent with this, do you agree to create specific functions for the following methods that already include parameter flags?

Yes.

> Should the name be seeOrphanEventTriggered or seeOrphanedEventTriggered ?

What is an orphaned event?

TavoNiievez: It is also not documented on Symfony.com but that is how it was named when it was added to the Profiler.
See https://github.com/symfony/symfony/issues/24347 and https://github.com/symfony/symfony/pull/24392

TavoNiievez: Basically it is an event that fires without the need for a listener. Instead, it is executed by calling the method `->dispatch` manually.

PR #63, opened 2020-11-27
Updated Test Project

TavoNiievez:
Closes #54

Issue #54, opened 2020-11-23
Missing tests


TavoNiievez:
> Good. Once it is merged we should have corresponding tests in symfony-demo 
_Originally posted by @DavertMik in https://github.com/Codeception/module-symfony/issues/50#issuecomment-732076031_

Due to the current test coverage of the module and the time it would take to correct compatibility issues in current test projects, I decided to create a [new test project](https://github.com/TavoNiievez/codeception-symfony-tests) with 3 branches:
- `main`, for the latest version of Symfony (5.2)
- `4.4`, for the latest LTS and,
- `3.4` for the oldest LTS.

<hr/>

<details>
<summary><b>Click to see progress on Symfony 5.2</b></summary>
<br/>

- [x] amLoggedInAs
- [x] amOnAction
- [x] amOnRoute
- [x] dontSeeAuthentication
- [x] dontSeeEmailIsSent
- [x] dontSeeFormErrors
- [x] grabParameter
- [x] grabRepository
- [x] grabService
- [x] logout
- [x] runSymfonyConsoleCommand
- [x] seeAuthentication
- [x] seeCurrentActionIs
- [x] seeCurrentRouteIs
- [x] seeEmailIsSent
- [x] seeFormErrorMessage
- [x] seeFormHasErrors
- [x] seeInCurrentRoute
- [x] seeInSession
- [x] seeNumRecords
- [x] seePageIsAvailable
- [x] seePageRedirectsTo
- [x] seeUserHasRole
- [x] seeUserPasswordDoesNotNeedRehash
- [x] submitSymfonyForm
</details>

<hr/>

<details>
<summary><b>Click to see progress on Symfony 4.4</b></summary>
<br/>

- [x] amLoggedInAs
- [x] amOnAction
- [x] amOnRoute
- [x] dontSeeAuthentication
- [x] dontSeeEmailIsSent
- [x] dontSeeFormErrors
- [x] grabParameter
- [x] grabRepository
- [x] grabService
- [x] logout
- [x] runSymfonyConsoleCommand
- [x] seeAuthentication
- [x] seeCurrentActionIs
- [x] seeCurrentRouteIs
- [x] seeEmailIsSent
- [x] seeFormErrorMessage
- [x] seeFormHasErrors
- [x] seeInCurrentRoute
- [x] seeInSession
- [x] seeNumRecords
- [x] seePageIsAvailable
- [x] seePageRedirectsTo
- [x] seeUserHasRole
[...]

PR #69, opened 2020-12-09
Remove PHP 7.0 support

TavoNiievez:
In accordance with what is mentioned in (#39) this module proceeds to progressively remove support for obsolete versions of PHP.

This time 7.0, which has not had any kind of support for 2 years.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "minimum-stability": "RC",
     "require": {
-        "php": ">=7.0.8 <9.0",
+        "php": ">=7.1.3 <9.0",
         "codeception/lib-innerbrowser": "^1.3",
         "codeception/codeception": "^4.0",
         "ext-json": "*"


Naktibalda:
Are you going to remove support for PHP 7.1 in a month as well?

TavoNiievez:
I would like to clarify that i did not randomly wait a month to send this PR.

What i actually did after removing support for `PHP 5.6` with (#39) was remove all the legacy code for `PHP 5.6` (#51) and include the `PHP 7.0` features where possible (#41).

**The intention is** to remove support from all unsupported versions of PHP. Which are: `PHP 7.0`, `PHP 7.1` and `PHP 7.2`. But it is not convenient to remove the support for all of them in a single PR.

First, an unsupported version is removed and then, all the features of the new minimum supported version are added. It has to be done little by little.

What happened is that a more critical task [emerged](https://github.com/Codeception/module-symfony/pull/50#issuecomment-732076031) and i had to prioritize it (#54) (#63).

So the answer is no, it won't be in a month, it may be less or it may be more, it depends on how long it takes me to update the code base, but **the goal is** to only have support for `PHP 7.3`, `PHP 7.4` and `PHP 8.0`.

Naktibalda:
Alright, it makes sense.

It would look better if legacy support was removed and/or features of newer PHP vere used in the same PR.

TavoNiievez:
> It would look better if legacy support was removed and/or features of newer PHP vere used in the same PR.

I thought of that,
So far i was doing it in separate PR because it may be that the functions that i or someone else are adding can benefit from those changes, or that those changes can create conflicts, or behavioral changes in those functions.

But now we have tests on all of them, and i'll try to do it in a single PR with `PHP 7.1` and `PHP 7.2` and as much as I can, thanks.

PR #77, opened 2020-12-12
Fix time collector logic when debugging

TavoNiievez:
I looked in the [history of the TimeDataCollector](https://github.com/symfony/http-kernel/blob/5.x/DataCollector/TimeDataCollector.php) and it never got called `'timer'` as well as it never had a method called `->getTime()` so i guess this never worked before.

I also added the missing docblock.


DavertMik (approved):

DavertMik:
Thank you, good catch!

PR #93, opened 2021-01-07
Minor changes to contribution guides

ThomasLandauer:


TavoNiievez:
@ThomasLandauer I am also [proposing changes to the test project readme](https://github.com/Codeception/symfony-module-tests/pull/9/files), your opinion on this is appreciated.

The part about using composer instead of the current steps is still tentative and can be discarded, but the description would be nice to include anyway.

TavoNiievez (approved):

PR #98, opened 2021-01-11
Added info from the 'Codeception for Symfony' page

ThomasLandauer:
I'd like to drop all those pages completely:
* https://codeception.com/for/symfony
* https://codeception.com/for/laravel
* etc.

I'm starting with Symfony (to show it's possible), and hopefully can motivate others to do the same for the other frameworks :-)

Reasons:
* Three pages explaining how to use Codeception with Symfony (each differently) are two too much :-) https://codeception.com/docs/modules/Symfony and https://codeception.com/for/symfony and https://codeception.com/09-04-2015/using-codeception-for-symfony-projects.html
  So the module page should be the only one, and it should contain everything.
* "use cases" is a misleading title for it; in fact it's just about frameworks ;-) WordPress and Joomla are exceptions - so if it's not possible to integrate them elsewhere, they can stay (renaming the button to "CMS")

So what I did is: I copied the "Features" list from https://codeception.com/for/symfony into the module page - and that's it! The rest of https://codeception.com/for/symfony has (a) nothing to do with Symfony (e.g. the shown configuration of "Acceptance Testing"), or (b) is explained elsewhere (e.g. BDD).

Questions:
* By "Symfony Crawler" you mean "DomCrawler"? (double-check the link I added)
* How does this work: "Symfony Router can be cached between requests to speed up testing"?
* I'm not 100% sure about the "API Tests" header, so please double-check if anything important from https://codeception.com/for/symfony#api-tests is missing at https://codeception.com/docs/10-APITesting#REST-API

So if you agree, I'll go ahead and delete https://github.com/Codeception/codeception.github.com/blob/master/for/symfony.md. If possible, the URL should be redirected to the module page.

BTW: What is https://github.com/Codeception/codeception.github.com/blob/master/for/zend-framework.md%7E ?

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -77,7 +77,14 @@ use function strpos;
 use function substr_compare;
 
 /**
- * This module uses Symfony Crawler and HttpKernel to emulate requests and test response.
+ * This module uses [Symfony's DomCrawler](https://symfony.com/doc/current/components/dom_crawler.html)
+ * and [HttpKernel Component](https://symfony.com/doc/current/components/http_kernel.html) to emulate requests and test response.
+ *
+ * * Access Symfony services through the dependency injection container: [`$I->grabService(...)`](#grabService)
+ * * Use Doctrine to test against the database: `$I->seeInRepository(...)` - see [Doctrine Module](https://codeception.com/docs/modules/Doctrine2)
+ * * Assert that emails would have been sent: [`$I->seeEmailIsSent()`](#seeEmailIsSent)
+ * * Tests are wrapped into Doctrine transaction to speed them up.
+ * * Symfony Router can be cached between requests to speed up testing.
  *
  * ## Demo Project
  *


TavoNiievez:
both **Naktibalda** and **DavertMik** have previously mentioned that they prefer to have the documentation and source code in one place, it's more likely to be up-to-date and in-sync that way, so you've got my 👍🏾 too.

> By "Symfony Crawler" you mean "DomCrawler"? (double-check the link I added)

Yes. Your link is ok.

> How does this work: "Symfony Router can be cached between requests to speed up testing"?

This is a very good question, in fact, for me that question could be an issue, since there is no documentation on how `$I->invalidateCachedRouter()` works or when exactly I should use it (neither does it have any test or example).

Here:
https://github.com/Codeception/module-symfony/blob/776a68977417f3470faf6eb2544e2280328d5cff/src/Codeception/Module/Symfony.php#L116
there is even reference to an article where optimizations are made in Sylius, but it is not very clear exactly how this solution:
```php
	if ($this->config['cache_router'] === true) {
		$this->persistPermanentService('router');
	}
```
was inferred.

TavoNiievez:
As an added bonus, currently on a fresh installation of Symfony:
```
symfony new newProject
```
If the first thing I do is install the Symfony module:
```
composer require codeception/module-symfony --dev
```
I run into several additional steps before I can run my first test.

The first one is that the Codeception recipe includes by default 3 suites with the Asserts and PhpBrowser modules activated. So:
```
composer require codeception/module-phpbrowser --dev
composer require codeception/module-asserts --dev
```
Then the initial steps don't tell me that I need to run:
```
vendor/bin/codecept build
vendor/bin/codecept run
```
Nor do they give me a link to the documentation.

Finally, [the first thing I get](https://github.com/Codeception/module-symfony/blob/3da97ac40369ff3e2942859e639911cdd14bdfd9/src/Codeception/Module/Symfony.php#L407) when executing any of the assertions that depend on the profiler is:
```
fail: The Profile is needed to use the 'xyz' function.
```
Without telling me which command to execute or taking me to the documentation about the Profiler.

as I see it, there are many steps to get started with the module.

Naktibalda:
I think that For page was supposed to be more of a marketing piece than technical reference.

Zend Framework page has never been finished.

ThomasLandauer:
Well, OK, if everybody agrees, then let's go! :-)

Steps:

1. Merge this.
2. Redirect the URL https://codeception.com/for/symfony to the module page. Or is this configured at some place where I can send a PR?
3. Delete the page from the repo (I can send a PR for this).

@TavoNiievez about the `cache_router`: I don't know anything about this. But in any case, this question shouldn't block this PR.

TavoNiievez:
> **TavoNiievez** about the cache_router: I don't know anything about this. But in any case, this question shouldn't block this PR.

@ThomasLandauer I only answered your question, I agree that it is not an impediment for this PR.

Naktibalda:
For step 2, add HTML redirection tag or javascript script to https://github.com/Codeception/codeception.github.com/blob/master/for/symfony.md

Related change would be to remove Symfony from Use Cases dropdown in https://codeception.com/
https://github.com/Codeception/codeception.github.com/blob/master/_includes/for.html

ThomasLandauer:
Please merge this PR now - the information should be present on Symfony Module page anyway - no matter how far the "deletion process" of the "use cases" dropdown already is...

ThomasLandauer:
What about https://codeception.com/for/joomla ? At first sight, most stuff is explained at https://github.com/joomla-projects/joomla-browser anyway. We could add a "CMS" header at https://codeception.com/docs/03-AcceptanceTests and  collect everything there?
Same is probably true for https://codeception.com/for/wordpress
Should I ask the maintainers?

Naktibalda:
A problem with adding more information to 03-AcceptanceTests or other numbered documentation pages is that these pages are already very long, so you are effectively burying these links and making AcceptanceTests page even more difficult to read.

There are links to joomla-browser and wp-browser in Addons page: https://codeception.com/addons

I think that the point of adding Wordpress and Joomla pages to separate top level menu was to highlight that Codeception works with these popular framework.

ThomasLandauer:
Yes, indeed :-) I already suggested to shorten them https://github.com/Codeception/Codeception/issues/4394 - but that's quite some work :-)

Anyway, what can we do at short notice?

* "burying" is indeed the right term, but this also applies to the "use cases" dropdown we're talking about ;-) So where needs really important stuff to be shown? => At the homepage! And - lo and behold - Wordpress and Joomla are already there :-) (see below for details).
* After thinking some more about it: Let's not make the same mistake again, and "copy" the same content to 7 different pages (of which 6½ get outdated by the time :-) So in this case: Mention it at a prominent place (i.e. home page - done), but for details **refer to their own websites**

Suggestions for the home page https://codeception.com/ (could be moved into a separate issue)
* Shorten the gray `<div>`, to make the next section ("Benefits_") rise up higher. I almost missed this, since I didn't scroll down at the beginning...
* General question: Which message are you sending out to Drupal users if you keep repeating "WordPress" and "Joomla" over and over? In other words: Are those WordPress and Joomla addons really adding something *substantial*, or just one or two convenience functions? I *guess* the correct message is not "Codeception works with WordPress and Joomla", but rather: "Codeception works with *any* CMS. For WordPress and Joomla there's an addon to make it even easier/better."
* It the frameworks "card" (bottom) replace all links leading to "for"-pages to the module page. For WordPress and Joomla link directly to their GitHub repo.
* Those "cards" and the "Framework_Testing" block (in the middle) are not in sync (e.g. Zend Framework missing), and some are missing completely (e.g. Laminas :-)

Where is this generated? Can I send a PR?

DavertMik:
@ThomasLandauer codeception for symfony and codeception for laravel pages exist mostly for maketing reasons. They are not really docs.
I am against removing them without any good alternative. 

If you know how to bring more Symfony/Laravel users into Codeception - the ideas are appreciated.

PR #100, opened 2021-01-13
Update the `services` Part

TavoNiievez:
Just to be consistent, the `services` part is updated, which is the only one documented (and useful?).

The `email` part was never documented (or used?), so that excess annotation is removed as well.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -114,7 +114,11 @@ use function sprintf;
  *
  * ## Parts
  *
- * * `services`: Symfony dependency injection container (DIC)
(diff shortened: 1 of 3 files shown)


Naktibalda:
I guess you have a good answer to question why we need parts.

The idea for exposing a few methods via parts was to make them available when Symfony application is being tested using PhpBrowser or WebDriver.

Most of your new parts only make sense for functional testing using Symfony module.
Using those parts would hide all methods provided by InnerBrowser.

TavoNiievez:
@Naktibalda So...

Would it be okay if I modify the PR so that only the `services` part remains?

PR #101, opened 2021-01-14
Refactor email assertions

TavoNiievez:
In this PR:

- Support for Swift Mailer is removed.
- The `mailer:` config parameter is removed since it is no longer needed.
- Now it is possible to assert emails sent from Symfony commands.
- The **Profiler** is replaced by Mail Logger to do the email assertions since the profiler requires 'requests' to be made and that is not always a requirement to send emails in Symfony.

The fewest possible changes were made to the API to make it easier for users to upgrade to version 2.

The most significant change, however, is that `seeEmailIsSent` now defaults to `1` instead of `null`. This change is made because previously an assert of type `assertGreatherThan` was made, whereas now, the `EmailCount Constraint` requires that the expected number of emails be specified.

Checking the exact amount of emails sent is probably what you want to assert in your tests.

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -36,6 +36,7 @@ use Symfony\Component\HttpKernel\DataCollector\TimeDataCollector;
 use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\HttpKernel\Profiler\Profile;
 use Symfony\Component\HttpKernel\Profiler\Profiler;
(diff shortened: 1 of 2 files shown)


TavoNiievez:
@ThomasLandauer please review this.

This fixes your issue https://github.com/Codeception/Codeception/issues/4698 so I am complying with what is mentioned in https://github.com/Codeception/module-symfony/pull/24#issuecomment-714809600

PR #106, opened 2021-01-25
Update code standards

TavoNiievez:


src/Codeception/Module/Symfony.php
```diff
@@ -340,14 +337,15 @@ protected function getKernelClass(): string
...
             );
         }
 
-        if (file_exists(codecept_root_dir() . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php')) {
-            // ensure autoloader from this dir is loaded
-            require_once codecept_root_dir() . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php';
+        $autoLoader = codecept_root_dir() . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php';
+
+        if (file_exists($autoLoader)) {
+            require_once $autoLoader;
```
Naktibalda, line 348: This code is odd, I think its only purpose was to get CI working, because it runs `composer install` in two separate directories.

TavoNiievez: Yes, it seems that it is, and it is still necessary... if I remove it I get:
`
PHP Fatal error:  Uncaught Error: Class 'Symfony\Component\HttpKernel\Kernel' not found in /home/runner/work/module-symfony/module-symfony/framework-tests/src/Kernel.php:12
`
So I opted to encapsulate it in a separate method improving its documentation:

```php
    /**
     * Ensures autoloader loading of additional directories.
     * It is only required for CI jobs to run correctly.
     */
    private function requireAdditionalAutoloader(): void
    {
        $autoLoader = codecept_root_dir() . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php';
        if (file_exists($autoLoader)) {
            require_once $autoLoader;
        }
    }
```

PR #119, opened 2021-03-15
Adding @var annotation for return type

ThomasLandauer:
... to better emphasize what you'll get back; especially since `@return \Symfony\Component\Mime\Email|null` isn't rendered right, due to https://github.com/Codeception/Codeception/issues/6104

So I added a docblock within the docblock, so to say ;-)


TavoNiievez:
This implicitly and explicitly is a solution for unwanted rendering behavior on the website.
However, this is not the ideal solution.

Besides, you can't add this annotation because you close the comment block `*/` and this causes the CI build to fail.

You can change `john_doe@user.com` to `john_doe@example.com` in the other PR that you have open.

PR #120, opened 2021-03-15
Added "See also"

ThomasLandauer:
I shortened the example, cause `$emails[0]->getTo()[0];` looks strange ;-) Instead, I added the "See also"-links which (I think) are a good idea in general since the page gets longer and longer, so it's harder to get an overview what functions do exist at all...

The `@var` annotation syntax is taken from https://phpstan.org/writing-php-code/phpdoc-types#general-arrays - I'm not 100% sure if this is the right way.

The closing `*/` in my new `@var` line now breaks the code completely ;-) (same as in https://github.com/Codeception/module-symfony/pull/119) - is there a way to escape this?

### DIFF
--- a/src/Codeception/Module/Symfony/MailerAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/MailerAssertionsTrait.php
@@ -39,6 +39,7 @@ trait MailerAssertionsTrait
 
     /**
      * Returns the last sent email.
+     * See also: [grabSentEmails()](https://codeception.com/docs/modules/Symfony#grabSentEmails)
      *
      * ```php
      * <?php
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony/MailerAssertionsTrait.php
```diff
@@ -61,12 +62,12 @@ public function grabLastSentEmail(): ?Email
 
     /**
      * Returns an array of all sent emails.
+     * See also: [grabLastSentEmail()](https://codeception.com/docs/modules/Symfony#grabLastSentEmail).
      *
      * ```php
      * <?php
+     * /** @var array<int, \Symfony\Component\Mime\Email> $emails */
```
TavoNiievez, line 69: Please remove this @var annotation, you can't `*/` in the middle of a docblock

TavoNiievez (changes requested):

src/Codeception/Module/Symfony/MailerAssertionsTrait.php
ThomasLandauer: But I like the idea - it makes the code sample so nicely copy-pastable :-)
I now tried to escape it - according to https://stackoverflow.com/a/1439351/1668200
Let's see how this comes out - otherwise I'll change it back...

TavoNiievez: I would instead only change:
```diff
- * @return \Symfony\Component\Mime\Email[]
+ * @return array<int, \Symfony\Component\Mime\Email>
```

TavoNiievez:
@ThomasLandauer please resolve conflicts.

ThomasLandauer:
I started from scratch ;-)

PR #124, opened 2021-03-16
Standardize "@example.com" as domain in documentation

ThomasLandauer:
`example.com` is the "official" domain for such things: http://example.com/

For the removal, see https://github.com/Codeception/module-symfony/pull/120#issue-593067918

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -110,7 +110,7 @@ use function sprintf;
  *         - Doctrine2:
  *             depends: Symfony
  *         - WebDriver:
- *             url: http://your-url.com
+ *             url: http://example.com
  *             browser: firefox
  * ```
  *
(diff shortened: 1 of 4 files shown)


TavoNiievez:
If you want to standardize @ example.com as the official domain for examples that's fine with me, however, doing a bit of Ctrl + F on the documentation page I can see that:

- `amLoggedInAs` has `'john_doe@gmail.com'`,
- `fillField` has `'jon@mail.com'`, (this one comes from InnerBrowser, so you'd have to send a PR there to change it).
- `grabSentEmails` has `'john_doe@user.com'`,
- `submitSymfonyForm` has `'john_doe@gmail.com'`.

Please update them all to be able to approve and merge this PR.

ThomasLandauer:
There you go - I hope I got 'em all :-)

BTW: For changes like this, it would be really handy to have a direct link from each function on the docs page to the source code, as I suggested at https://github.com/Codeception/Codeception/discussions/6063#discussioncomment-244542
What do you think? Who could add this? @Arhell too?

TavoNiievez:
> For changes like this, it would be really handy to have a direct link from each function on the docs page to the source code

I know what you mean. But the symfony module is currently the only one that has traits for functions, so this change would only be for this module.

Previously only this:

imagen

was enough because all the documentation was in the same file, but this is no longer the case.

> Who could add this? @Arhell too?

Are you in the Slack of Codeception? Sadly I don't have an answer for that either.

TavoNiievez (approved):

Issue #125, opened 2021-03-22
Status of Swift Mailer support in v2.0.0?


ThomasLandauer:
@TavoNiievez Am I right that the `mailer` configuration you/we introduced in https://github.com/Codeception/module-symfony/pull/9/files was removed in v2.0.0, and the module just doesn't work with Swift Mailer anymore?

At https://github.com/Codeception/module-symfony/releases/tag/2.0.0 you only mentioned that Symfony 4.4 is required, but this doesn't automatically mean that Swift Mailer was dropped in favor of Symfony Mailer, cause Swift Mailer is still supported in Symfony 4.4: https://symfony.com/doc/4.4/email.html

This is the "Fail" message people get when running their tests:

> Emails can't be tested without Symfony Mailer service.

And I'd say it should be reworded into something like:

> `codeception/module-symfony` requires Symfony Mailer to test emails. So you need to either upgrade your app from Swift Mailer to Symfony Mailer https://symfony.com/doc/current/mailer.html or downgrade `codeception/module-symfony` to 1.6.0

... cause some people might not have heard of Symfony Mailer at all yet ;-)

But it would be better to inform people *upfront* that it's not going to work. How could we do that? Maybe a yellow box at https://codeception.com/docs/modules/Symfony somewhere near the top? I can come up with some PR, but want to hear your opinion first :-)

BTW: On that page, what is that blue "table" in the top right: Changelog | Source | 3.1 | 2.5 | 1.8 ? The version numbers look like Codeception's, so they're quite misleading here. Do you know how/where this can be changed?

TavoNiievez:
@ThomasLandauer As always, thanks for opening.

You're right.

I updated [the release notes](https://github.com/Codeception/module-symfony/releases/tag/2.0.0) to include that explanation. Is it okay as it is?

> And I'd say it should be reworded (the "Fail" message)

Yes, you're right, it seems like something important. The idea is that if you use Swift Mailer and you don't want (or can't) update to Symfony Mailer just stay on the 1.x branch of this module, please send a PR... the message you mention is fine for me.

> Do you know how/where this can be changed? (the blue "table" in the top right)

It seems to come from [this file](https://github.com/Codeception/codeception.github.com/blob/1568f32c874a3eb6fc48b5ea2448c4dfca73cf84/docs/modules/Symfony.md). That part shouldn't be overwritten when generating the documentation automatically, as far as I know.

TavoNiievez:
[...]

PR #126, opened 2021-03-24
Mentioning Symfony Mailer requirement in v2.0.0

ThomasLandauer:
Follow-up of https://github.com/Codeception/module-symfony/issues/125. My plan would be to copy this line to all email-related functions. Questions:
1. Is `^1.6` right?
2. The `mailer` configuration was never documented anywhere. But in v1.6, Swift Mailer is the default anyway, right? So this configuration is only relevant for people who still have v1.6 but use Symfony Mailer? And in this case, the better advice is to upgrade to v2, rather than configuring `mailer`?

### DIFF
--- a/src/Codeception/Module/Symfony/MailerAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/MailerAssertionsTrait.php
@@ -15,6 +15,7 @@ trait MailerAssertionsTrait
      * Checks that no email was sent.
      * The check is based on `\Symfony\Component\Mailer\EventListener\MessageLoggerListener`, which means:
      * If your app performs a HTTP redirect, you need to suppress it using [stopFollowingRedirects()](https://codeception.com/docs/modules/Symfony#stopFollowingRedirects) first; otherwise this check will *always* pass.
+     * Starting with version 2.0.0, `codeception/module-symfony` requires your app to use [Symfony Mailer](https://symfony.com/doc/current/mailer.html). If your app still uses [Swift Mailer](https://symfony.com/doc/current/email.html), set your version constraint to `^1.6`.
      */
     public function dontSeeEmailIsSent(): void
     {
@@ -43,6 +44,7 @@ trait MailerAssertionsTrait
      * Returns the last sent email.
      * The function is based on `\Symfony\Component\Mailer\EventListener\MessageLoggerListener`, which means:
      * If your app performs a HTTP redirect after sending the email, you need to suppress it using [stopFollowingRedirects()](https://codeception.com/docs/modules/Symfony#stopFollowingRedirects) first.
+     * Starting with version 2.0.0, `codeception/module-symfony` requires your app to use [Symfony Mailer](https://symfony.com/doc/current/mailer.html). If your app still uses [Swift Mailer](https://symfony.com/doc/current/email.html), set your version constraint to `^1.6`.
      * See also: [grabSentEmails()](https://codeception.com/docs/modules/Symfony#grabSentEmails)
      *
      * ```php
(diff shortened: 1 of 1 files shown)


TavoNiievez:
Do you think we can move this to the top of the documentation page ?, I'm not sure if this is the best place to remember something like that.

TavoNiievez:
> Is ^1.6 right?

Not necessarily, any version 1.x is compatible with Swift Mailer, the most recent (1.6) is the most recommended.

> The mailer configuration was never documented anywhere.

Yes, [it was](https://github.com/Codeception/module-symfony/pull/9/commits/80b7c3229d07dee168d5aad316cb1f44018e0e76) documented.

> in v1.6, Swift Mailer is the default anyway, right?

Yes.

> So this configuration is only relevant for people who still have v1.6 but use Symfony Mailer?

Yes, this includes people who use unsupported versions of Symfony, from 4.0 to 4.3, since these versions are not officially supported of any kind, it does not make sense to support them in the module.

> And in this case, the better advice is to upgrade to v2, rather than configuring mailer?

Yes.

ThomasLandauer:
> Do you think we can move this to the top of the documentation page ?, I'm not sure if this is the best place to remember something like that.

I thought about that too. My counter-arguments were:
* It's not important for *everybody* but just for users of the email functions.
* If somebody misses it, the fix is quite easy (downgrade), so it's not really an "alert" we need to throw upon everybody.
* If it's mentioned at the very function which ultimately causes the fail, it probably gets more attention (by the "relevant" people) than if it's included in the "foreword".

But I'm not 100% sure myself ;-) What/where would you suggest?
By "remember" to you mean the users, or you/us (the maintainers)?

TavoNiievez:
> By "remember" to you mean the users, or you/us (the maintainers)?

I mean the users, not the maintainers.

> What/where would you suggest?

After reading your own counterarguments, I still think it's best to just say it once.

ThomasLandauer:
> After reading your own counterarguments, I still think it's best to just say it once.

OK, then some box at the top with the current text. But on each function a link to this box?! (In the end, the link will be almost as long as the entire text... ;-)

TavoNiievez:
> But on each function a link to this box?!

No, no link.

Having to mention compatibility with specific versions of other packages in the method docblocks is something I don't like at all.

I think a better solution would be to nip the problem in the bud and directly block the installation of version 2.0 if you have swift mailer installed.

As far as I know, composer is smart enough in those scenarios to install the closest lower version that does not include a [conflicts](https://getcomposer.org/doc/04-schema.md#conflict) directive.

Or I understand that Codeception has a "ConflictsWith" interface to handle these cases.

@Naktibalda What do you think?

Naktibalda:
Codeception has ConflictsWithModule interface which is not applicable in this case.

`conflicts` in composer.json looks like a good solution, but you would have to re-release 2.0.0 with added conflicts clause to get it working properly.
A small number of users who already have 2.0.0 installed and use SwiftMailer won't be happy about it.

TavoNiievez:
Ok, then I will do the following:

1. I will let @ThomasLandauer add these messages in each method as a form of warning.
2. I will re-release 2.0.0 with all the changes planned for 2.0.1 including this PR (which, so far include only documentation changes, it will not interfere with anyone's tests).
3. I will immediately send a PR adding the composer conflict directive and removing all the added warning messages and release it as version 2.0.1.

ThomasLandauer:
> Having to mention compatibility with specific versions of other packages

Which other packages? We're only talking about *this module's versions.* The current text in this PR is valid "forever" (so no maintenance problem).

But more important: @TavoNiievez I don't think your plan is a good solution. Reasons:
* Having Swift Mailer *installed* does not conflict with v2! You just can't use the email related functions ;-)
* On large projects, moving away from Swift Mailer might be done in steps. If they start migrating the important emails to Symfony Mailer first, they can use v2, and still continue to use Swift Mailer for some not-so-important emails (for which they don't have tests).

So my suggestions are (in that order):
1. Back to my original plan (i.e. repeat the warning on each function). I mean, there's so many outdated information on this website (which I'm trying to get rid of, step by step), that the worst case scenario (a handful of *irrelevant* - not outdated! - notices on some functions) doesn't really threaten me ;-)
2. If you still don't like that, let's go for a caution box on top of the page, with links from each function.
3. If you don't like that either, just the caution box (no links)
4. If you don't like that too, just close this PR and do nothing. Sooner or later they'll see the Fail message which explains it anyway ;-)

But this entire question is not important enough for ongoing discussion! Whichever solution it'll gonna be, the time spent discussing it is already longer than the time it'll take to implement it ;-)
So pick one: 1,2,3,4

TavoNiievez:
In my previous comment I already mentioned that you can implement your **1.** (in this PR) for version `2.0.0` (which I agreed to be re-released).

I will keep a message at the top (without links) (i.e. your **3.**) for `2.0.1` so that it can be viewed on the documentation page.

The rest has already been discussed.

TavoNiievez (approved):

TavoNiievez:
@ThomasLandauer Thank you for your patience in this matter.

ThomasLandauer:
@TavoNiievez Never mind :-)
But I still think that doing a `2.0.1` release is not a good idea! (see reasons above...)

PR #128, opened 2021-04-04
submitSymfonyForm(): Mentioning `name` attribute

ThomasLandauer:
I'm not sure about this!
But I think the submitted fieldname now is `myFormaddress`, instead of `myForm[address]` - that's what I'm trying to fix.
I didn't take a look at the tests you mentioned in https://github.com/Codeception/module-symfony/issues/54

BTW: I was first looking for this method in `FormAssertionsTrait.php`, is there a reason why you put it into `BrowserAssertionsTrait.php`?

### DIFF
--- a/src/Codeception/Module/Symfony/BrowserAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/BrowserAssertionsTrait.php
@@ -83,7 +83,7 @@ trait BrowserAssertionsTrait
     /**
      * Submit a form specifying the form name only once.
      *
-     * Use this function instead of `$I->submitForm()` to avoid repeating the form name in the field selectors.
+     * Use this function instead of [`$I->submitForm()`](#submitForm) to avoid repeating the form name in the field selectors.
      * If you customized the names of the field selectors use `$I->submitForm()` for full control.
      *
      * ```php
--- a/src/Codeception/Module/Symfony/BrowserAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/BrowserAssertionsTrait.php
@@ -103,7 +103,7 @@ trait BrowserAssertionsTrait
 
         $params = [];
         foreach ($fields as $key => $value) {
-            $fixedKey = sprintf('%s[%s]', $name, $key);
+            $fixedKey = sprintf('%s%s', $name, $key);
             $params[$fixedKey] = $value;
         }
         $button = sprintf('%s_submit', $name);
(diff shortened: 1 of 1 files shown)


TavoNiievez:
As shown in the code example,

```php
$I->submitSymfonyForm('login_form', [
     '[email]'    => 'john_doe@example.com',
 	 '[password]' => 'secretForest'
]);
```
You need to add brackets to the key of the field that is passed as argument (eg. `'[first_name]'`). The reason behind this is that there are more 'complex' forms that might contain compound fields, for example:

```php
[password][first] => 'mysupersecretpass'
[password][second] => 'mysupersecretpass'
```

and at the time I thought it was better to leave the bracket syntax than to invent a special syntax just to send compound fields to this function.

> BTW: I was first looking for this method in FormAssertionsTrait.php, is there a reason why you put it into BrowserAssertionsTrait.php?

yes, the functions that are in BrowserAssertionTrait don't really use any framework-specific Symfony class. Neither of forms nor of anything else.

What is the same, they could be integrated in Innerbrowser and be available in all modules.
I did not send a PR to Innerbrowser for two reasons:

- I think that the Innerbrowser module, despite working perfectly well, needs a little extra work to update its code base and include new PHP features, as well as a refactor in general since it is difficult in my opinion to navigate [through that file](https://github.com/Codeception/lib-innerbrowser/blob/master/src/Codeception/Lib/InnerBrowser.php). 
But that module affects all modules at the same time and any change of that style can be difficult to merge. (Too many people affected, too few reviewing).

- The methods of that trait were ideas that I came up with to make some tests much easier to write and read, but knowing if they are being used as expected or there are difficulties (or redundancies) when using it is a feedback that I can only get later that it merges and that some people use that function in open source projects (and I can review them). For example, that drove the changes I made with [seePageIsAvailable](https://github.com/Codeception/module-symfony/pull/114).

The idea of ​​doing this is to be very sure that both the API and the documentation are well designed and written, that it is understood and that they are used by people when writing their tests.

So my hope is that in some future those functions can be copied to InnerBrowser and removed from this module.

TavoNiievez:
btw, don't forget to tag me when you ask questions directed at me, I am not the only maintainer of this module :-)

ThomasLandauer:
Sorry, I overlooked the brackets ;-)
You can merge this now.

TavoNiievez (approved):

TavoNiievez:
Thank you.

Issue #66, opened 2020-12-02
Rename all abbreviated function arguments to better support PHP8's Named Arguments


ThomasLandauer:
If I'm understanding [PHP8's Named Arguments](https://www.php.net/releases/8.0/de.php?lang=de#named-arguments) right, instead of
```php
$I->seeInSession('name', 'Thomas');
```
...you can now do
```php
$I->seeInSession(value: 'Thomas', attrib: 'name');
```
The problem I'm seeing is that some functions have abbreviated/slang parameter names like `params` or `val` (instead of `parmeters` or `value`). If we change them **now**, it's no real BC break, since nobody (?) is using this feature yet. I'm suggesting this for the Symfony module first, to get started with a smaller "sub-project". https://github.com/Codeception/module-symfony/pull/65 is a first step.

Questions:

1. Do you agree in principle that all occurrences of `params` should be renamed to `parameters`?
2. Do you think that it's worth trying to do it automatically? Scrolling through `Symfony.php` I don't see *that* many occurrences, so doing it manually is probably faster in total?

TavoNiievez:
> Do you agree in principle that all occurrences of params should be renamed to parameters?

<details>
<summary>Clic to see answer.</summary>
<br/>

**Disclaimer:** the following is just my *personal* opinion:

I disagree. changing 'params' to 'parameters' is just a style change. There is really no standard that says that to better support named arguments they must contain whole words. In fact, in the default php functions you see these abbreviations very often.

So I think this would be a trivial change.

What I could agree with is to be congruent and if 'params' is used, it is used throughout the module. But for this you would also have to change this in the other modules (and in all the code of Codeception also). Some of the functions that i included simply -copied- the nomenclature of the parameters in Laravel module to try to create a unified API between modules. The effort it would take to send PRs to all modules to change the abbreviations **(or keep them)** just don't think it compensates, or means a real objective improvement in readability.
Unless Codeception had defined a naming guide that everyone must follow, doing this would not be very useful, because in the future anyone could propose exactly the opposite of what you propose now and no one could actually contradict it.
</details>

TavoNiievez:
Also note that the following functions also have abbreviations in their parameter names:
```php
public function amOnRoute(string $routeName, array $params = [])
public function seeCurrentRouteIs(string $routeName, array $params = [])
public function seeNumRecords(int $expectedNum, string $className, array $criteria = [])
public function amOnAction(string $action, array $params = [])
```

[...]

PR #132, opened 2021-04-24
Added seeRequestTimeIsLessThan function

TavoNiievez:
Code Example:

```php
	$I->amOnPage('/dashboard');
	$I->seeRequestTimeIsLessThan(40.5);
```

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -21,6 +21,7 @@ use Codeception\Module\Symfony\RouterAssertionsTrait;
 use Codeception\Module\Symfony\SecurityAssertionsTrait;
 use Codeception\Module\Symfony\ServicesAssertionsTrait;
 use Codeception\Module\Symfony\SessionAssertionsTrait;
+use Codeception\Module\Symfony\TimeAssertionsTrait;
 use Codeception\Module\Symfony\TwigAssertionsTrait;
 use Codeception\TestInterface;
 use Exception;
@@ -138,6 +139,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
     use SecurityAssertionsTrait;
     use ServicesAssertionsTrait;
     use SessionAssertionsTrait;
+    use TimeAssertionsTrait;
     use TwigAssertionsTrait;
 
     /**
--- a/src/Codeception/Module/Symfony/TimeAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/TimeAssertionsTrait.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Codeception\Module\Symfony;
+
+use Symfony\Component\HttpKernel\DataCollector\TimeDataCollector;
+use function number_format;
+use function sprintf;
+
+trait TimeAssertionsTrait
+{
+    /**
+     * Asserts that the time a request lasted is less than expected (`$expectedTime`).
+     *
+     * @param float $expectedTime The expected time in milliseconds
+     */
+    public function seeRequestElapsedTimeLessThan(float $expectedTime): void
+    {
+        $timeCollector = $this->grabTimeCollector(__FUNCTION__);
+
+        $realTime = $timeCollector->getDuration();
+
+        $this->assertLessThan(
+            $expectedTime,
+            $realTime,
+            sprintf(
+                'The request was expected to last less than %s ms, but it actually lasted %s ms.',
+                number_format($expectedTime, 2),
+                number_format($realTime, 2)
+            )
+        );
+    }
+
+    protected function grabTimeCollector(string $function): TimeDataCollector
+    {
+        return $this->grabCollector('time', $function);
+    }
+}
--- a/src/Codeception/Module/Symfony/TimeAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/TimeAssertionsTrait.php
@@ -5,13 +5,13 @@ declare(strict_types=1);
 namespace Codeception\Module\Symfony;
 
 use Symfony\Component\HttpKernel\DataCollector\TimeDataCollector;
-use function number_format;
+use function round;
 use function sprintf;
 
 trait TimeAssertionsTrait
 {
     /**
-     * Asserts that the time a request lasted is less than expected (`$expectedTime`).
+     * Asserts that the time a request lasted is less than expected.
      *
      * If the page performed a HTTP redirect, only the time of the last request will be taken into account.
      * You can modify this behavior using [stopFollowingRedirects()](https://codeception.com/docs/modules/Symfony#stopFollowingRedirects) first.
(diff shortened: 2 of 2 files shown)


TavoNiievez:
@ThomasLandauer

Naktibalda:
The big question is do we want to add performance testing capability to Codeception.
The big problem that comes with it is that code coverage makes PHP much slower, so it would be a good idea to disable this function if code coverage is turned on.

I think that method name could be improved, but I hope that @ThomasLandauer can offer some better names.

The choice of time unit is non-obvious and it would be better to tell what it is in the method name, e.g. `seeRequestElapsedTimeInMillisecondsIsLessThan`.

What is the expected behaviour of this assertion if `/dashboard` page redirects using Location header - measure execution time of the last request or of all requests made since `amOnPage`?

TavoNiievez:
> The big question is do we want to add performance testing capability to Codeception.

It is not my intention to add a whole layer of utilities and assertions for that. But I think it would be nice if I at least have an assertion for that, if I don't want to install a specific package for performance testing.

> The big problem that comes with it is that code coverage makes PHP much slower, so it would be a good idea to disable this function if code coverage is turned on.

For this specific case, that does not apply. I use Symfony's TimeDataCollector which is already collecting that information. All I do is get it and make an assertion with it.

To disable it, and speed up the test or code coverage, you would have to disable or uninstall the Profiler, which would make most of the assertions of this module stop working.

> I think that method name could be improved

Me too :)

> The choice of time unit is non-obvious and it would be better to tell what it is in the method name, e.g. seeRequestElapsedTimeInMillisecondsIsLessThan.

I thought about that, but it would also be strange to talk about request time in seconds, for example. In the majority of cases they do not usually take that long. However I agree that it would be even better if the name of the method directly gives me that information... (if it didn't get too long in the process)

> What is the expected behaviour of this assertion if /dashboard page redirects using Location header - measure execution time of the last request or of all requests made since amOnPage?

Only the time of the last request is taken into account (this seems relevant enough to be added to the docblock).

Naktibalda:
> To disable it, and speed up the test or code coverage, you would have to disable or uninstall the Profiler, which would make most of the assertions of this module stop working.

My point wasn't that your assertion slows down code coverage, it was another way around.  For example, if some request takes 35 milliseconds in normal circumstances and you set your assertion to 40 milliseconds to get warning if it becomes slower, with code coverage, request would take a few times longer, e.g. 100 milliseconds and your assertion would fail.

TavoNiievez:
oh yeah, you're right.

But I'm not sure; I wouldn't want Codeception to disable my assertions automatically by default in those cases.

I would prefer if I explicitly have to add a configuration in my suite.yml file if I want, for example, that if I have code-coverage activated, mark those tests as incomplete (`markTestIncomplete`) or directly ignore them (`markTestSkipped`) .

If you enlighten me on how to know if the code coverage is active I could write that code.

Naktibalda:
There is no way to know if code coverage is active in Codeception, but PHPUnit 10 introduced static method for that.

ThomasLandauer:
No matter how, but it would be nice if the final result was shorter than `seeHttpRequestWasFasterThanTheLimitWhichOurManagementDidSetToTheFollowingNumberOfMilliseconds()` ;-)

OK, back to work :-)

* Unit inside the name is (a) too long and (b) very uncommon - with (b) probably being a consequence of (a). So +1 for the idea of communicating it in the argument's name (which will get more important in the future due to PHP 8). But I'd say that `$ms` is enough here, cause that says it all.
* "Only the time of the last request is taken into account" This means that redirects are not counted? Anyway, I would rather not include "Latest" in the function name, since (a) redirects are a bad practice anyway ;-), and (b) they usually don't take long, so it makes not much of a difference - right? So I'd say explaining this in the docblock is enough.
* "Duration" doesn't add any value over "Time" - it just sounds more complicated ;-) And "time" is also the term used in Symfony's web toolbar & profiler.
* Most existing functions do contain a verb, even if they are long (e.g. `seeResponseCodeIsClientError`)

So in total, I'm suggesting:
**seeRequestTimeIsLessThan(ms:50)**

Just about "Request" I'm not 100% convinced. Is "request time" a commonly used term? What about "ResponseTime"?
What does it actually measure? The time the server takes for internal processing, until it starts sending the HTTP string through the wire?

@TavoNiievez I think what Naktibalda means here:

> The big problem that comes with it is that code coverage makes PHP much slower, so it would be a good idea to disable this function if code coverage is turned on.

... is that we/you should not try to solve this in the code, but rather just *tell* people (in the docs) to manually turn off code coverage. @Naktibalda right?

TavoNiievez:
> he time the server takes for internal processing, until it starts sending the HTTP string through the wire?

As far as I know what `getDuration()` returns is the time it takes to resolve a request. That is, request time + response time, hence the '*Elapsed*'.

btw, are you sure of `$ms`? Sometimes you say that abbreviations *in general* are bad practice...
(In that case, `$realTime` also needs to be renamed.)

Naktibalda:
Parameter name doesn't make test code longer, so it could be changed to $milliseconds or $expectedMilliseconds or even $expectedTimeInMilliseconds :)

 >  > The time the server takes for internal processing, until it starts sending the HTTP string through the wire?

 > As far as I know what getDuration() returns is the time it takes for the server to resolve (internally) a request, hence the 'Elapsed'.

It is time [since kernel start](https://github.com/symfony/symfony/blob/v5.2.6/src/Symfony/Component/HttpKernel/DataCollector/TimeDataCollector.php#L41-L45), so I think that this assertion could be affected by setting [rebootable_client parameter](https://codeception.com/docs/modules/Symfony#Config) to false.

ThomasLandauer:
> btw, are you sure of $ms? Sometimes you say that abbreviations in general are bad practice...

Yeah, you caught me here ;-) I suggested this at https://github.com/Codeception/module-symfony/issues/66
The difference to me is: Things like "attribs" and "em" are programmers' slang. "ms" is the "official" short version of this unit, like "kg".

> Parameter name doesn't make test code longer

That's certainly right for today. But due to PHP 8, the best practices might change, and in 5 years it could be commonplace to write `foo(length: 7)`. Don't you think?
However, it's not that important - so if you prefer, take `$milliseconds` ;-)

ThomasLandauer:
BTW: Is it really necessary to use a `float` (instead of `int`) for the milliseconds?

BTW2: I would omit "expected" from `$expectedMilliseconds`, since it's clear that you cannot pass the *actual* milliseconds ;-)

TavoNiievez:
> Is it really necessary to use a float (instead of int) for the milliseconds?

`getDuration()` returns float, it makes sense to compare it to another float. There is no need to restrict the decimal point in this case.

> BTW2: I would omit "expected" from $expectedMilliseconds, since it's clear that you cannot pass the actual milliseconds ;-)

> > But due to PHP 8, the best practices might change, and in 5 years it could be commonplace to write foo(length: 7)

I kept that in mind, however, the main use of named arguments at the moment is to allow arguments to be passed in a different order than defined in the function. Most of the cases when there are things like this: `$this->someMethod([], [], [], [], null, nulll, $myArgument)`. So, I think it can stay that way for now.

TavoNiievez:
> It is time since kernel start, so I think that this assertion could be affected by setting rebootable_client parameter to false.

I'm going to test this, if it does affect then I'll add it to the docblock.

ThomasLandauer:
> getDuration() returns float, it makes sense to compare it to another float.

But then, if you do `seeRequestTimeIsLessThan(50)`, PhpStan will give you a warning about the wrong type, right? So maybe you could accept `float|int` and then do something like:
```php
$ms  = is_int($ms) ? (float) $ms : $ms;
```

Or (even better) cast the output of `getDuration()` to `int`, cause nobody will ever setup a limit in *fractions* of milliseconds ;-)

TavoNiievez:
okay, allowing integers as well as floats does sound good. I will make the change soon.

TavoNiievez:
Effectively, that's right.
If you have `rebootable_client: false` set, this assertion stops working if the page performs redirects or if you access multiple pages using `amOnPage` (or similar) in a row:

```
Fail The Profile is needed to use the 'seeRequestTimeIsLessThan' function.
```

Any ideas for documentation of this behavior?

ThomasLandauer:
> Any ideas for documentation of this behavior?

My suggestion is https://github.com/TavoNiievez/module-symfony/pull/1 :-)

TavoNiievez:
Thanks @ThomasLandauer :)
Ok I think this is ready to merge.

PR #134, opened 2021-05-29
Added logout by path

TavoNiievez:
Now that a simpler interface for logging out [is being discussed in Symfony](https://github.com/symfony/symfony/issues/40663), I think it's a good time to clarify that there are two ways to log out in a functional test.

One is by going the logout path,
and the other is by invalidating the session and expiring the cookies 'manually' or _'programmatically'_.

### DIFF
--- a/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
@@ -10,6 +10,7 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGenerator;
 use function is_int;
 use function serialize;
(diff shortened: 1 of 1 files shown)


TavoNiievez:
@ThomasLandauer

ThomasLandauer:
Sorry, I don't quite get the benefit here :-( If I want to test the app's logout feature, I certainly have to use your first method (`$I->amOnPage('/logout')`).

So what would be a use case for a programmatic logout in a test? If I wanted to proceed with a logged out (=not logged in) user, I'd just start a new test ;-)

TavoNiievez:
@ThomasLandauer you can have an event subscriber for the logout event. This would fire when accessing `'/logout'`. `logoutProgramatically` is a way to clean up that session from Codeception and not from Symfony as such.
Another advantage is that the `/logout` path is customizable. If you use `$I->amOnPage` you would have to edit the test in case that path changes.

ThomasLandauer:
Sorry, I still don't really see a use case :-(

> `logoutProgramatically` is a way to clean up that session from Codeception and not from Symfony as such.

Yeah, but having Codeception and Symfony out of sync is a bad thing, isn't it?

> Another advantage is that the `/logout` path is customizable. If you use `$I->amOnPage` you would have to edit the test in case that path changes.

Yeah, OK, but that's just how life is ;-) If you change the app, you have to adapt the tests.

PR #136, opened 2021-05-31
Update readme.md

Arhell:


readme.md
```diff
@@ -22,6 +22,8 @@ composer require "codeception/module-symfony" --dev
 
 See [the module documentation](https://codeception.com/docs/modules/Symfony).
 
+[Changelog](https://github.com/Codeception/module-symfony/tags)
```
TavoNiievez, line 25: I think a user would find the [Releases](https://github.com/Codeception/module-symfony/releases) tab more useful than the Tags tab if they are looking to see a changelog.

Arhell: excellent remark, I will do

PR #137, opened 2021-06-11
Support for Codeception 5

TavoNiievez:


TavoNiievez:
This requires more work, I'll open a PR when it's ready again on its respective branch.

PR #145, opened 2021-09-27
Improving error message for `public` services

ThomasLandauer:
Is `\n` correct for a line break?


TavoNiievez (approved):
The line break is fine. 

Thanks!

Arhell (approved):

PR #149, opened 2021-12-07
Update dependencies

ThomasLandauer:
I'm just guessing this - so if it's wrong, just close it :-)

Same situation is in module-phpbrowser: https://github.com/Codeception/module-phpbrowser/blob/master/composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,12 +18,12 @@
     "require": {
         "php": "^7.4 | ^8.0",
         "ext-json": "*",
-        "codeception/lib-innerbrowser": "^1.4|^2.0",
-        "codeception/codeception": "^4.0"
+        "codeception/lib-innerbrowser": "^2.0",
+        "codeception/codeception": "^4.1"
     },
     "require-dev": {
-        "codeception/module-asserts": "^1.3",
-        "codeception/module-doctrine2": "^1.1",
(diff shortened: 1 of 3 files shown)


TavoNiievez:
Hi @ThomasLandauer , eventually there should be a release of `module-symfony` (and any other modules) using the versions released today.
However, you will most likely not use both (1.x and 2.x) at the same time.

Because one module depends on another, and so on, updating all of them at once proved a bit difficult. But without a doubt this is something that will be done in the next version.

I haven't yet verified `module-symfony` compatibility with Symfony 5.4 and 6.0 and I even have new features pending, so thanks for pointing out that here is something else to add to the to-do list!

Btw, you can also try to fix the build yourself and we can merge this before, if you want. I suggest you first update the dependency locally, PHPStorm should help you by marking incompatibilities in red.

TavoNiievez:
I wanted to avoid mixing updating PHP version with updating dependencies so as not to do a super difficult update, but you are right, it is something that should be done.

ThomasLandauer:
Never mind, it's OK for me, just thought you were maybe missing something...
All good! :-)

TavoNiievez:
thank you very much :-) I will try to do it as soon as possible

TavoNiievez:
@ThomasLandauer thanks you :)

PR #154, opened 2022-02-09
[Codecept5] Improved support for Codeception 5/PHP 8

yesdevnull:
I need to do the following:

- [x] Consolidate Symfony version requirements compared to Codeception 5 (e.g. supported PHP versions).
Symfony 4.4 and 5.4 are LTS versions so are [updated to support the latest PHP versions](https://symfony.com/doc/current/contributing/community/releases.html#php-compatibility).

- [ ] Fix `codeception/codeception` version requirement.
- [ ] Fix `codeception/lib-innerbrowser` version requirement.
- [ ] Add several basic smoke tests?
- [ ] Correct commit message to be consistent with repository.

Bonus points (for a separate PR maybe):
- [ ] PHPStan (using config from [`module-doctrine2`](https://github.com/Codeception/module-doctrine2/blob/master/phpstan.neon))

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,13 +27,13 @@
         "codeception/module-asserts": "^2.0 || *@dev",
         "codeception/module-doctrine2": "^2.0 || *@dev",
         "doctrine/orm": "^2.10",
-        "symfony/form": "^4.4 | ^5.0 | ^6.0",
-        "symfony/framework-bundle": "^4.4 | ^5.0 | ^6.0",
-        "symfony/http-kernel": "^4.4 | ^5.0 | ^6.0",
-        "symfony/mailer": "^4.4 | ^5.0 | ^6.0",
-        "symfony/routing": "^4.4 | ^5.0 | ^6.0",
-        "symfony/security-bundle": "^4.4 | ^5.0 | ^6.0",
-        "symfony/twig-bundle": "^4.4 | ^5.0 | ^6.0",
+        "symfony/form": "^4.4 || ^5.4 || ^6.0",
+        "symfony/framework-bundle": "^4.4 || ^5.4 || ^6.0",
+        "symfony/http-kernel": "^4.4 || ^5.4 || ^6.0",
+        "symfony/mailer": "^4.4 || ^5.4 || ^6.0",
+        "symfony/routing": "^4.4 || ^5.4 || ^6.0",
+        "symfony/security-bundle": "^4.4 || ^5.4 || ^6.0",
+        "symfony/twig-bundle": "^4.4 || ^5.4 || ^6.0",
         "vlucas/phpdotenv": "^4.2 || ^5.3"
     },
     "suggest": {
@@ -47,6 +47,7 @@
         ]
     },
     "config": {
-        "classmap-authoritative": true
+        "classmap-authoritative": true,
+        "sort-packages": true
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,12 @@
 {
     "name": "codeception/module-symfony",
     "description": "Codeception module for Symfony framework",
-    "keywords": ["codeception", "symfony"],
-    "homepage": "https://codeception.com/",
-    "type": "library",
     "license": "MIT",
+    "type": "library",
+    "keywords": [
+        "codeception",
+        "symfony"
+    ],
     "authors": [
         {
             "name": "Michael Bodnarchuk"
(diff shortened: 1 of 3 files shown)


yesdevnull (commented):
I normalised the composer.json file using [ergebnis/composer-normalize](https://github.com/ergebnis/composer-normalize). I like the consistency of a normalised composer.json file but happy to revert if you'd prefer it to be kept as-is.

TavoNiievez:
@yesdevnull The next thing to do is to update [lib-innerbrower](https://github.com/Codeception/lib-innerbrowser) to be compatible with the changes from `Codeception/dev-5.0-interfaces`.

This module doesn't support Symfony 6 yet... although i already added the corresponding branch in the test project https://github.com/codeception/symfony-module-tests/tree/6.0 .

Which leads me to say that we don't need smoke tests as we **do** have tests for this. Please check the aforementioned [codeception-module-tests](https://github.com/Codeception/symfony-module-tests), maybe it should be mentioned in the `README.md` that we have the tests in another repository.
It can help you understand how it works by reading the [main.yml](https://github.com/Codeception/module-symfony/blob/main/.github/workflows/main.yml) file and maybe [CONTRIBUTING.md](https://github.com/Codeception/module-symfony/blob/main/CONTRIBUTING.md)

I'm a big fan of all the initiatives to 'normalize', 'update' and 'standardize', so I support you on the topic of normalizing the `composer.json` file.
The only thing is that since this is normalization, it would only make sense if all of Codeception uses the same normalized style. So I would ask you to raise PR to update that in all of our repositories if you want to go ahead.

Lastly, regarding PHPStan, yes please!
A separate PR would be nice.

composer.json
```diff
@@ -14,32 +16,35 @@
...
-        "symfony/twig-bundle": "^4.4 | ^5.0",
-        "vlucas/phpdotenv": "^4.2 | ^5.3"
+        "symfony/form": "^4.4 | ^5.0 | ^6.0",
+        "symfony/framework-bundle": "^4.4 | ^5.0 | ^6.0",
+        "symfony/http-kernel": "^4.4 | ^5.0 | ^6.0",
+        "symfony/mailer": "^4.4 | ^5.0 | ^6.0",
+        "symfony/routing": "^4.4 | ^5.0 | ^6.0",
+        "symfony/security-bundle": "^4.4 | ^5.0 | ^6.0",
+        "symfony/twig-bundle": "^4.4 | ^5.0 | ^6.0",
+        "vlucas/phpdotenv": "^4.2 || ^5.3"
```
TavoNiievez, line 37: let's use simple | everywhere.

yesdevnull: As mentioned [here](https://github.com/Codeception/module-doctrine2/pull/55#discussion_r804464812), Composer recommend using `||` over `|`. I'm inclined to follow their example, but up to you 😄

TavoNiievez: Well, what the documentation says is that it will continue to support the simple version for compatibility. There is no mention that it is deprecated or that it will be removed in future versions.

I personally used the simple one since, well, it's simpler. But I have no problem using double since we are changing the style of the file anyway.

readme.md
```diff
@@ -9,7 +9,7 @@ A Codeception module for Symfony framework.
 
 ## Requirements
 
-* `Symfony 4.4` or higher.
+* `Symfony` `4.4.x`, `5.4.x` or `6.x` or higher, as per the [Symfony supported versions](https://symfony.com/releases).
```
TavoNiievez, line 12: well done!

composer.json
yesdevnull: > There is no mention that it is deprecated or that it will be removed in future versions.

That is very true. I suspect it'll never be deprecated but the language of the note suggests that it's recommended to use `||` instead (at least that's my reading of it).

composer.json
```diff
@@ -14,34 +16,37 @@
...
-    "minimum-stability": "dev",
+    "homepage": "https://codeception.com/",
     "require": {
-        "php": "^7.4 | ^8.0",
+        "php": "^8.0",
         "ext-json": "*",
-        "codeception/lib-innerbrowser": "^2.0 | *@dev",
-        "codeception/codeception": "dev-5.0-interfaces as 5.0.x-dev"
+        "codeception/codeception": "dev-5.0-interfaces as 5.0.0",
+        "codeception/lib-innerbrowser": "dev-master"
```
Naktibalda, line 24: dev-5.0-interfaces branch is gone, you can set codeception/codeception to `^5.0.0-alpha2` and lib-innerbrowser to `^3.0`.

PR #161, opened 2022-09-05
Improving language

ThomasLandauer:


Naktibalda:
Thanks,

CI failed in an interesting way, I will try to fix it later.

PR #167, opened 2023-02-06
Allow installation of codeception/lib-innerbrowser v4

W0rma:
v4 added support for PHPUnit v10 (https://github.com/Codeception/lib-innerbrowser/releases/tag/4.0.0)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php": "^8.0",
         "ext-json": "*",
         "codeception/codeception": "^5.0.0-RC3",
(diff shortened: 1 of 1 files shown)


Naktibalda:
Thanks, I completely forgot about this.

Naktibalda:
Github action fails because PHPUnit 9 and  PHPUnit 10 are loaded at the same time, probably https://github.com/Codeception/symfony-module-tests installs PHPUnit 9. I will release a new version anyway,

PR #173, opened 2023-12-11
Add seeEvent/dontSeeEvent event assertions

xEdelweiss:
Hi! This is a follow-up for #168. In this update, I've added the `seeEvent`/`dontSeeEvent` methods, which check if an event was dispatched without considering whether the listener processed it. 

This approach allows the assertion to cover both orphan and non-orphan events. Therefore, this should help keep tests "green" during development, especially in scenarios where listeners might be added after the event's introduction.

Additionally, this PR consolidates `assertEventTriggered` and `assertEventNotTriggered` into a single method to reduce duplication, as the method became a bit more complex.

[PR with tests.](https://github.com/Codeception/symfony-module-tests/pull/25)

### DIFF
--- a/src/Codeception/Module/Symfony/EventsAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/EventsAssertionsTrait.php
@@ -39,7 +39,40 @@ trait EventsAssertionsTrait
         if ($expected === null) {
             $this->assertSame(0, $data->count());
         } else {
-            $this->assertEventNotTriggered($data, $expected);
+            $this->assertEventTriggered($data, $expected, true);
+        }
+    }
+
+    /**
+     * Verifies that there were no events during the test.
+     * Orphan events are included. Use `dontSeeOrphanEvent` to exclude them.
+     *
+     * ```php
+     *  <?php
+     *  $I->dontSeeEvent();
+     *  $I->dontSeeEvent('App\MyEvent');
+     *  $I->dontSeeEvent(new App\Events\MyEvent());
+     *  $I->dontSeeEvent(['App\MyEvent', 'App\MyOtherEvent']);
+     *  ```
+     *
+     * @param array|object|string|null $expected
+     */
+    public function dontSeeEvent(array|object|string $expected = null): void
+    {
+        $eventCollector = $this->grabEventCollector(__FUNCTION__);
+
+        $data = [
+            $eventCollector->getOrphanedEvents(),
+            $eventCollector->getCalledListeners(),
+        ];
+        $expected = is_array($expected) ? $expected : [$expected];
+
+        if ($expected === null) {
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony/EventsAssertionsTrait.php
```diff
@@ -39,7 +39,40 @@ public function dontSeeOrphanEvent(array|object|string $expected = null): void
...
             $this->assertSame(0, $data->count());
         } else {
-            $this->assertEventNotTriggered($data, $expected);
+            $this->assertEventTriggered($data, $expected, true);
+        }
+    }
+
+    /**
+     * Verifies that there were no events during the test.
+     * Orphan events are included. Use `dontSeeOrphanEvent` to exclude them.
```
TavoNiievez, line 48: Can you rewrite this comment block in a way that makes it a little more clear what it means that this assertion 'includes' orphan events and in which cases `dontSeeOrphanEvent` should be used instead?
Although I understand your code it is difficult for me to understand the use of these terms in the comment block.
The same applies for `seeEvent`*

src/Codeception/Module/Symfony/EventsAssertionsTrait.php
```diff
@@ -177,33 +237,38 @@ public function seeEventListenerIsCalled(
...
             $expectedEvent = is_object($expectedEvent) ? $expectedEvent::class : $expectedEvent;
-            $this->assertTrue(
-                $this->eventWasTriggered($actual, (string)$expectedEvent),
-                "The '{$expectedEvent}' event did not trigger"
-            );
+            $message = $assertTrue
+                ? "The '{$expectedEvent}' event did not trigger"
+                : "The '{$expectedEvent}' event triggered";
+
+            $condition = false;
```
TavoNiievez, line 261: Could you use a more descriptive name for this variable?

TavoNiievez (changes requested):

src/Codeception/Module/Symfony/EventsAssertionsTrait.php
xEdelweiss: The idea was that `seeEvent` would let you know that the event was triggered, but not that it is non-orphan, so you need an additional call to `dontSeeOrphanEvent`. But then I messed it up while cleaning up.

I've removed "exclusion" from `dontSeeEvent` phpDoc as it cannot be applied there. And fixed the suggested method for `seeEvent`.

src/Codeception/Module/Symfony/EventsAssertionsTrait.php
xEdelweiss: Done. Hope it makes variable names clearer.

TavoNiievez (approved):

PR #182, opened 2024-02-04
#165 login token

### LINKED ISSUE #165: Token for amLoggedInAs should be configurable
I am testing some functionality regarding login and logout with an external OpenID provider. For that, I depend on some attributes that are set on the token.
In Symfony Tokens can have so called _attributes_. See `TokenInterface` (https://github.com/symfony/security-core/blob/6.2/Authentication/Token/TokenInterface.php#L76).

When using this modules `amLoggedInAs` method, I cannot modify the returned token,
[This](https://github.com/Codeception/module-symfony/blob/main/src/Codeception/Module/Symfony/SessionAssertionsTrait.php#L51) token generation should be switched to an approach, where I can control the token, f. ex. setting some attributes on the token.

And so my logic, that depends on attributes for the token fails.

codedge:
Closes #165

### DIFF
--- a/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
@@ -43,7 +43,7 @@ trait SessionAssertionsTrait
         $this->loginWithToken($token, $firewallName, $firewallContext);
     }
 
-    private function loginWithToken(TokenInterface $token, string $firewallName = 'main', string $firewallContext = null): void
+    protected function loginWithToken(TokenInterface $token, string $firewallName, ?string $firewallContext): void
     {
         $this->getTokenStorage()->setToken($token);
 
--- a/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/SessionAssertionsTrait.php
@@ -7,6 +7,7 @@ namespace Codeception\Module\Symfony;
 use Symfony\Component\BrowserKit\Cookie;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken;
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony/SessionAssertionsTrait.php
```diff
@@ -32,12 +33,21 @@ trait SessionAssertionsTrait
...
         $token = $this->createAuthenticationToken($user, $firewallName, $roles);
+        $this->loginWithToken($token, $firewallContext, $firewallName);
+    }
+
+    public function amLoggedInWithToken(TokenInterface $token, string $firewallName = 'main', string $firewallContext = null): void
+    {
+        $this->loginWithToken($token, $firewallName, $firewallContext);
+    }
+
+    private function loginWithToken(TokenInterface $token, string $firewallName = 'main', string $firewallContext = null): void
```
TavoNiievez, line 46: This function should be `protected` because the inheritance of the `Codeception\Module\Symfony` class is allowed and this method encapsulates a useful independent functionality,
on the other hand, unlike `amLoggedInAs` and `amLoggedInWithToken` none of its parameters should have default values because its invocation already carries a default parameter.

TavoNiievez (changes requested):

TavoNiievez:
There are errors unrelated to this PR in the GitHub Actions.
When I have some time I'll see how I can fix them.
For the time being, I would appreciate if you could add a test for this new method in the module test project [symfony-module-tests](https://github.com/codeception/symfony-module-tests) (See the [contribution guide](https://github.com/Codeception/module-symfony/blob/main/CONTRIBUTING.md)).

Thanks!

src/Codeception/Module/Symfony/SessionAssertionsTrait.php
codedge: Make sense, applied.

src/Codeception/Module/Symfony/SessionAssertionsTrait.php
```diff
@@ -32,12 +33,21 @@ trait SessionAssertionsTrait
      */
     public function amLoggedInAs(UserInterface $user, string $firewallName = 'main', string $firewallContext = null): void
     {
-        $session = $this->getCurrentSession();
         $roles = $user->getRoles();
-
         $token = $this->createAuthenticationToken($user, $firewallName, $roles);
+        $this->loginWithToken($token, $firewallContext, $firewallName);
```
TavoNiievez, line 38: The order of the arguments does not seem to match the signature of the method below,
the tests fail because of:

> [TypeError] Codeception\Module\Symfony::loginWithToken(): Argument 2 ($firewallName) must be of type string, null given, called in .../SessionAssertionsTrait.php on line 38

TavoNiievez (changes requested):

src/Codeception/Module/Symfony/SessionAssertionsTrait.php
TavoNiievez: @codedge tests are running again, so work on this RP can continue.

TavoNiievez: @codedge tests are running again, so work on this PR can continue.

TavoNiievez: I fixed this myself, but there are still a few things I'm going to double check before merging this.

PR #189, opened 2024-03-26
Added Symfony validator assertions

TavoNiievez:
Examples:
```php
        $user = User::create('test@example.com', 'password', ['ROLE_ADMIN']);

        $I->dontSeeViolatedConstraint($user);
        $I->dontSeeViolatedConstraint($user, 'email');
        $I->dontSeeViolatedConstraint($user, 'email', Email::class);
        $I->dontSeeViolatedConstraint($user, 'password', NotBlank::class);

        $user->setPassword('weak');
        $I->seeViolatedConstraint($user, 'password', Length::class);
```

```php
        $user = User::create('invalid_email', 'weak', ['ROLE_ADMIN']);

        $I->seeViolatedConstraintsCount(2, $user);
        $I->seeViolatedConstraintsCount(1, $user, 'email');
        $I->seeViolatedConstraintMessage('is not a valid email', $user, 'email');

        $user->setEmail('');
        $I->seeViolatedConstraintMessage('should not be blank', $user, 'email');
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
         "symfony/security-csrf": "^5.4 | ^6.4 | ^7.0",
         "symfony/security-http": "^5.4 | ^6.4 | ^7.0",
         "symfony/twig-bundle": "^5.4 | ^6.4 | ^7.0",
+        "symfony/validator": "^5.4 | ^6.4 | ^7.0",
         "symfony/var-exporter": "^5.4 | ^6.4 | ^7.0",
         "vlucas/phpdotenv": "^4.2 | ^5.4"
     },
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -24,6 +24,7 @@ use Codeception\Module\Symfony\ServicesAssertionsTrait;
 use Codeception\Module\Symfony\SessionAssertionsTrait;
 use Codeception\Module\Symfony\TimeAssertionsTrait;
 use Codeception\Module\Symfony\TwigAssertionsTrait;
+use Codeception\Module\Symfony\ValidatorAssertionsTrait;
 use Codeception\TestInterface;
 use Doctrine\ORM\EntityManagerInterface;
 use Exception;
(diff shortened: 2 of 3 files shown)


TavoNiievez:
@cs278 @W0rma @xEdelweiss @ThomasLandauer

These assertions verify that the Symfony validator component is working correctly in the application, specifically with the test data defined in the test cases.

If you have some free time please review my PR :-)
Thanks!

ThomasLandauer:
What about shortening the name a bit, from `dontSeeViolatedConstraint` to just `dontSeeViolation`?
And I think the "count" method should be plural: seeViolation**s**Count

TavoNiievez:
Hi @ThomasLandauer 
Indeed, I considered `seeViolation` and `seeViolated` but I think that when reading it is not explicit with what 'violation' I am referring to, so I wanted to avoid uncomfortable laughs or unintentional offenses in the developers who read this code.

For the moment, I agree that the assertion count should be plural, thank you very much for pointing that out.

ThomasLandauer:
IMO, "violation" is a common term in programming, so I guess it's politically correct :-) `dontSeeViolatedConstraint` is OK too, it's just that the method names are getting longer and longer...
What do the others say?

burned42:
If I think in terms of Symfony, validator constraints is a pretty commonly known naming, so I think my first guess reading `dontSeeViolation` would also be that it's about violation constraints from the validator component.
But that naming might not be that clear for everybody and I think using the naming violation constraint might be more explicit and if in doubt I tend to rather use longer naming than being ambiguous.
In the end, I don't have a strong preference here, because in Codeception I intentionally have to enable the Symfony module, so I'm aware of the Symfony context.

FYI, Symfony itself also doesn't use 'constraint' in the method naming, but then again, in the Symfony context that might not be necessary: https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Validator/Test/ConstraintValidatorTestCase.php

xEdelweiss:
Hi @TavoNiievez 
I am a little late, yet this is a great addition to the module!

I also like the shorter names, and the Symfony Serializer uses 'violations' as a key in normalization, so it should be easy to understand.

On the other hand, a longer name makes it easier to discover the method without looking at the docs. At least I try to "guess" the appropriate method first :) 
It's nice that you may find it by searching for either 'constraint' or 'violation'.

PR #190, opened 2024-04-16
Require bootstrap.php if exists, to load all necessary .env files

### LINKED ISSUE #99: Symfony: `.env.local` and `.env.test` are ignored
Starting in November 2018, Symfony has changed the way `.env` files are working: https://symfony.com/doc/current/configuration.html#the-env-file-environment-variables

These changes are not realized in Codeception.

To assert this, I created a controller which just outputs an environment variable:
```php
return new Response(getenv('MAILER_URL'));`
```

When I access the url with codeception (`codecept_debug($I->grabPageSource());`), I always get what I have in `.env` (should be `.env.local` or `.env.test`).

However, when I open the page in the browser in DEV environment, I get the content of `.env.dev`, when in TEST environment, `.env.test`; this is the expected behavior.

### Details

* Codeception version: 2.5.x-dev fc6deb8
* Symfony version: 4.2.1

TavoNiievez:
I'm replacing #4 with this.

In my last comment I expressed that I had problems with compatibility logic in lower versions of Symfony, but since the minimum compatible version is now 5.4 we can move on.
I require the `tests/bootstrap.php` file if it exists, otherwise I try to use dotEnv manually.
If all that fails I throw an exception telling the user what to do.

In the tests I performed this works regardless of whether or not you have `.env` and `.env.test` set in `codeception.yml`. 
This will also allow to remove these values from the [recipe](https://github.com/symfony/recipes-contrib/blob/8f56f2cee929025a5597b9e44d81fe90c4e7cd71/codeception/codeception/5.0/codeception.yml#L15) since they are now loaded automatically.

Closes #99.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "symfony/config": "^5.4 | ^6.4 | ^7.0",
         "symfony/dependency-injection": "^5.4 | ^6.4 | ^7.0",
         "symfony/dom-crawler": "^5.4 | ^6.4 | ^7.0",
+        "symfony/dotenv": "^5.4 | ^6.4 | ^7.0",
         "symfony/error-handler": "^5.4 | ^6.4 | ^7.0",
         "symfony/filesystem": "^5.4 | ^6.4 | ^7.0",
         "symfony/form": "^5.4 | ^6.4 | ^7.0",
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -86,6 +86,7 @@ use function sprintf;
  * * `cache_router`: 'false' - Enable router caching between tests in order to [increase performance](http://lakion.com/blog/how-did-we-speed-up-sylius-behat-suite-with-blackfire)
  * * `rebootable_client`: 'true' - Reboot client's kernel before each request
  * * `guard`: 'false' - Enable custom authentication system with guard (only for Symfony 5.4)
+ * * `bootstrap`: 'false' - Enable the test environment setup with the tests/bootstrap.php file if it exists or with Symfony DotEnv otherwise. If false, it does nothing.
  * * `authenticator`: 'false' - Reboot client's kernel before each request (only for Symfony 6.0 or higher)
  *
  * #### Sample `Functional.suite.yml`
@@ -169,6 +170,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         'em_service' => 'doctrine.orm.entity_manager',
         'rebootable_client' => true,
         'authenticator' => false,
+        'bootstrap' => false,
         'guard' => false
     ];
 
@@ -206,7 +208,9 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         }
 
         $this->kernel = new $this->kernelClass($this->config['environment'], $this->config['debug']);
-        $this->bootstrapEnvironment();
+        if($this->config['bootstrap']) {
+            $this->bootstrapEnvironment();
+        }
         $this->kernel->boot();
 
         if ($this->config['cache_router'] === true) {
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -28,11 +28,13 @@ use Codeception\Module\Symfony\ValidatorAssertionsTrait;
 use Codeception\TestInterface;
 use Doctrine\ORM\EntityManagerInterface;
 use Exception;
+use LogicException;
 use ReflectionClass;
 use ReflectionException;
 use Symfony\Bundle\SecurityBundle\DataCollector\SecurityDataCollector;
 use Symfony\Component\BrowserKit\AbstractBrowser;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface;
 use Symfony\Component\HttpKernel\DataCollector\TimeDataCollector;
@@ -204,6 +206,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         }
 
         $this->kernel = new $this->kernelClass($this->config['environment'], $this->config['debug']);
+        $this->bootstrapEnvironment();
         $this->kernel->boot();
 
         if ($this->config['cache_router'] === true) {
@@ -457,6 +460,26 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         }
 
         return array_unique($internalDomains);
+    }
+
+    private function bootstrapEnvironment(): void
+    {
+        $bootstrapFile = $this->kernel->getProjectDir() . '/tests/bootstrap.php';
+
+        if (file_exists($bootstrapFile)) {
+            require_once $bootstrapFile;
+        } else {
+            if (!method_exists(Dotenv::class, 'bootEnv')) {
+                throw new LogicException(
+                    "Symfony DotEnv is missing. Try running 'composer require symfony/dotenv'\n" .
+                    "If you can't install DotEnv add your env files to the 'params' key in codeception.yml\n" .
+                    "or update your symfony/framework-bundle recipe by running:\n" .
+                    'composer recipes:install symfony/framework-bundle --force'
+                );
+            }
+            $_ENV['APP_ENV'] = $this->config['environment'];
+            (new Dotenv())->bootEnv('.env');
+        }
     }
 
     /**


TavoNiievez:
@c33s I have updated the PR asking for `config/bootstrap.php` anyway.
The rest of the code you pasted is the same as what is being added in this PR.
Thanks!

c33s:
have you noticed my "edit"?

> edit: and yes you are right, the config/bootstrap.php is gone.

TavoNiievez:
Yes, I noticed it.

I also edited the description of the PR to indicate that, although it is no longer part of the recipe since 5.1, it is possible that someone has updated their project to 5.4 without updating the recipe, and because of Symfony's promise of backward compatibility it still works for them.

That's why I thought it was appropriate to ask about it, do you agree?

c33s:
i am a big fan of BC
the check for the file don't really hurt, i would keep it in. 

i just had a look at your code and thought about it. i am not sure if it is optimal. maybe it tries to think too much for "me".  you are loading the `tests/bootstrap.php` file. isn't the file already loaded by codeception by default? what happens if i define `bootstrap: _bootstrap.php` in `codeception.yml` won't i get a strange behavior if the symfony module tries to be smart and loads a hardcoded `test/bootstrap.php` file which may still exist but the developer switched to `tests/_bootstrap.php` via config and wonders which magic is happening.

if i want to load the `config/bootstrap.php` (for dotenv handling) but also want to use `test/bootstrap.php` (for loading `BypassFinals::enable();`) this would not be possible with this code.

i would want codeception to load the `tests/bootstrap.php` file from the codeception.yml (or whatever name was defined there) and the symfony module should only load the dotenv stuff via the old (config/bootstrap.php) or the newer method. 

also i am not sure if overriding the environment is the perfect solution but for that i am not sure.  

just some thoughts. what do you think? do i missunderstand something or are my thoughts valid?

TavoNiievez:
Ok.
What I understood from your previous message is 'hey, it would be nice if the bootstrap file to be loaded was the same as the one defined in the codeception.yml file'.
And well, I thought so too :-)

But at the time of writing the code and so far I don't know how to access that Codeception setting from the module code. It's probably something very simple, but before I figured it out I read the thread of the issue that this PR is targeting and most comments indicated that 'I prefer the Symfony module to handle the envs the same way Symfony does' and that is, through the tests/bootstrap.php file or from DotEnv.
If Codeception already loaded it before it won't do anything. That's why the require_once.
I understand that it would be cool if in addition to doing this by default it was also able to handle the file from the configuration to do all the customizations you want, but that's an improvement, not the solution to the issue, in other words "it's not a blocker".
If someone tells me how to access the data in the codecept.yml file I can make the modification without problems.

TavoNiievez:
By the way,
now that you mentioned it, I didn't know that Codeception generated a `tests/bootstrap.php` file.
and it seems wrong to me that it does since it seems to be based on the Symfony one, but if we look at the latest version of the recipe for phpunit-bridge:
https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/6.3/tests/bootstrap.php

the code has changed, and is likely to change again in the future.
It is not right to generate old versions of that file every time someone installs Codeception, nor that we have to update it manually every time it changes in Symfony.

c33s:
i think the symfony module should only 
- load the dotenv system 
- config/bootstrap.php if it exists
- don't interact with the `tests/bootstrap.php`  (shouldn't that be automatically loaded before through codeception? currently i load dotenv there and the symfony module has then access to all vars from the .env files) as far as i know this file is specific for codeception, there i can bootstrap my personal test environment (like loading bypass finals, set my environment, ...)
- should not hardcode the environment, sometimes i maybe have two test environments like `test` and `test-quick`

c33s:
i think you mix something up, we have `config/bootstrap` which is from legacy symfony, then we have the symfony php unit specific `tests/bootstrap.php` which is php-unit-symfony-bridge specific (has nothing to do with codeception, it only helps to integrate php unit with symfony, exactly what the bridge is made for). and then there is the "official" codeception `tests/bootstrap.php` this file is defined in `codeception.yml` and it used for booting up the codeception environment 
see https://codeception.com/docs/Customization > BOOTSTRAP section. it is just a naming/file collision between codeception and the phpunit symfony bridge.

TavoNiievez:
Ok
It is a very big coincidence that besides having the same name, they have the same code.
https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/5.1/tests/bootstrap.php
I think the version that Codeception should load when Symfony 4.4 is no longer supported is this one:
https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/5.3/tests/bootstrap.php

Anyway, answering your question, the module must interact with the tests/bootstrap.php file because otherwise, that file is not loaded automatically. The Codeception recipe defines the 'bootstrap' key and it is also not generated by default in new Codeception installations.

Regarding not 'hardcoding' the environment, I'm not hardcoding it, you must mandatorily define the environment in your module configuration file:
        - Symfony:
              app_path: 'src'
              environment: 'test
The value you define here is the value that will be taken into account in the tests, if you have another environment called differently, you will have to change it here, and it will be used with the code I added.

TavoNiievez:
It is a very big coincidence that besides having the same name, they have the same code.
https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/5.1/tests/bootstrap.php
I think the version that Codeception should load when Symfony 4.4 is no longer supported is this one:
https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/5.3/tests/bootstrap.php

Anyway, answering your question, the module must interact with the tests/bootstrap.php file because otherwise, that file is not loaded automatically. The Codeception recipe does not define the 'bootstrap' key and it is not generated by default in new Codeception installations.

Regarding not 'hardcoding' the environment, I'm not hardcoding it, you must mandatorily define the environment in your module configuration file:
        - Symfony:
              app_path: 'src'
              environment: 'test
The value you define here is the value that will be taken into account in the tests, if you have another environment called differently, you will have to change it here, and it will be used with the code I added.

c33s:
in gerneral i mean we have to look at codeceptions bootstrap.php in a more global perspective. we also have to assume the file exists without symfony. and even with symfony the files maybe completly different if i want to boot my test env differently. so i would disagree that the two files share the same code (even if they may).

with hardcoded i mean, that you will always overwrite a already set environment. for example if a set `test` in the config as my environment and then use codeceptions bootstrap.php to change this to `test-fast` with some custom code your code will always reset the environment to `test`. that was my definition of hardcoded. i can't switch the symfony environment dynamically any more.

> The module must interact with tests/bootstrap.php for automatic loading. 

can you explain that please? why must the module interact with it? isn't the file loaded by codeception automatically if i defined `bootstrap: bootstrap.php`. i think we are fixing the wrong problem. if the recipes  tests/boostrap.php files is not loaded we must change the default config generated by codeception to include `bootstrap: bootstrap.php`

another problem would be that your code will prevent to load codeceptions bootstap.php at all if config/bootstrap.php exists. why would you want that exactly? if i load bypass finals in codeceptions bootstrap.php and then want to boostrap symfony with config/bootstrap.php this would no longer work with your code. your code would require me to delete config/bootstrap.php and then manually add its code to codeceptions bootstrap.php so that it would work again. 

but i don't know codceptions internals enough, maybe we even don't need to load tests/bootstrap.php in the module at all (would make the most sense to me) as it was already loaded by codeception before. 
in this case symfony module should only look for config/bootstrap.php and load it if its there (then your code would work but we could remove the tests/bootstrap.php handling completly)

so my suggestion would be:
- verify how codeception interacts with its bootstrap.php (loaded before symfony module?) is the `bootstrap: bootstrap.php` the default config?
- handle the case that `tests/bootstrap.php` exists but only contains non-symfony boostraping which must be loaded.
- maybe even check if dotenv is already loaded
- only change the environment to the config value if it is not already set / or allow to disable the environment setting via module config.

does this make sense for you?

c33s (changes requested):
see conversation

c33s:
for reference from a fresh codeception installation:

`codeception.yml` (from codeception receipe):
```
namespace: App\Tests
support_namespace: Support
paths:
    tests: tests
    output: tests/_output
    data: tests/Support/Data
    support: tests/Support
    envs: tests/_envs
actor_suffix: Tester
extensions:
    enabled:
        - Codeception\Extension\RunFailed
params:
    - .env
    - .env.test
settings:
    shuffle: true
    colors: true
    report_useless_tests: true
```

`tests/bootstrap.php` (from phpunit bridge receipe): 
```
<?php

use Symfony\Component\Dotenv\Dotenv;

require dirname(__DIR__).'/vendor/autoload.php';

if (file_exists(dirname(__DIR__).'/config/bootstrap.php')) {
    require dirname(__DIR__).'/config/bootstrap.php';
} elseif (method_exists(Dotenv::class, 'bootEnv')) {
    (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
}
```

maybe we need to add change the manual for installing the symfony module. don't think adding to much magic really helps. looks like we will more likely run in unwanted sideeffects.

TavoNiievez:
> verify how codeception interacts with its bootstrap.php (loaded before symfony module?)
> is the bootstrap: bootstrap.php the default config?

I have told you before that no, it does not exist in new installations.
no, it is not part of the recipe either,
and no, if it does not exist it is not loaded before the symfony module, it is simply never loaded, the tests are never run because the environment is not loaded.
I am not guessing, I tested it before opening this PR. I invite you to test this behavior.
Even if you modify the recipe there will be thousands of people out there who don't have that key.

> handle the case that tests/bootstrap.php exists but only contains non-symfony boostraping which must be loaded.

"The user can modify the bootstrap.php file generated by Codeception/Symfony to remove the part where it loads environments and add their own custom code" is not something I'm going to fix in this PR. 

> maybe even check if dotenv is already loaded

It doesn't load.

> only change the environment to the config value if it is not already set / or allow to disable the environment setting via module config.

You cannot disable the environment setting. It is mandatory.
If you change the environment for some reason in the bootstrap file and not in the codeception.yml file your code will run after this setting and will still work.

I don't need to worry about loading `tests/bootstrap.php` if I already loaded `config/bootstrap.php` in this PR because
if you are expecting `tests/bootstrap.php` to do something custom you should have defined it in your `codeception.yml`, and if so it will load anyway.

Please before commenting further I suggest you do some testing of your own.
Conversations based on your concerns conjecturing about what might happen are getting us nowhere.

c33s:
we talk past one another and maybe you missunderstood me. i invite you to do a voice chat with me to clearify things.

i just argue for being able to have both: a custom bootstrap.php file without all symfony dotenv handling and the magic of the symfony module which would also load the dotenv code. this is not possible with your code.

TavoNiievez:
you can reach me via Slack

TavoNiievez:
> i just argue for being able to have both: a custom bootstrap.php file without all symfony dotenv handling and the magic of the symfony module which would also load the dotenv code. this is not possible with your code.

@c33s Done in the last commit. Anything else?

c33s:
@TavoNiievez 
no slack here. i switched to matrix.
i currently opended a jitsi meet here if you like: https://fairmeeting.net/MaximumTransportsInheritSlightly (will be online there the next 2 hours)

the config parameter is perfect. it would be enough for me but it is not what i meant. 

i meant to also be able to load tests/boostrap.php (custom without dotenv) **and** let the symfony module do its magic to do all the rest (so the default config settings should simply "make it work" tm.

to be completly flexible and allow all developers to customize their needs i would go more in a config-for-everyting direction.
```php
    private function bootstrapEnvironment(): void
    {
        // with that i can manage if i want to override an existing env
        if (empty($_ENV['APP_ENV']) || $this->config['set_app_env']) {
            $_ENV['APP_ENV'] = $this->config['environment'];
        }

        // so i can load config/bootstrap.php or tests/bootstrap.php or nothing or for example tests/symfonyExtraBootstrap.php
        if (!empty($this->config['bootstrap_file'])) {
            $bootstrapFile = $this->kernel->getProjectDir() . $this->config['bootstrap_file'];
            if (file_exists($bootstrapFile)) {
                require_once $bootstrapFile;
            }
        }

        if ($this->config['load_dotenv'] && !method_exists(Dotenv::class, 'bootEnv')) {
            throw new LogicException(
                "Symfony DotEnv is missing. Try running 'composer require symfony/dotenv'\n" .
                "If you can't install DotEnv add your env files to the 'params' key in codeception.yml\n" .
                "or update your symfony/framework-bundle recipe by running:\n" .
                'composer recipes:install symfony/framework-bundle --force'
            );
            (new Dotenv())->bootEnv('.env');
        }
    }
```

TavoNiievez:
Ok, I understand what you want to do, but again I reiterate, that would be an (very custom) enhancement.
And it's not a use case that this PR is solving.

This PR is dedicated to those who are having problems loading their env files just as they would expect them to be loaded in Symfony, nothing more, nothing less.

The `$_ENV['app_env']` assignment is only going to happen if you manually enable it in the `boostrap` module setting in the first place and decide to let Symfony DotEnv do the work instead of a bootstrap.php file.

If that's not what you need just don't enable this feature.

c33s:
it's sad that you block so simple config enhancements which can help developers because they don't sit in front of hardcoded values which they can't control. this only leads to duplicate code by reimplementing stuff which is already there in a fully custom boostrap file. it is sad that you build a cool feature that people which use `bootstrap: bootstrap.php` only can disable.

in many projects i worked on `tests/bootstap.php` is "owned" by codeception with custom code. symfony boostrap is loaded afterwards. 

i will leave the jitis meet in the next 10mins as i assume you won't come anyway.

my vote is clearly for more configs. 

over and out

TavoNiievez:
@c33s thank you for your help in this matter.

c33s:
at least leave my comment alone and unhide it

PR #199, opened 2024-09-19
Fix parameter name in exception for case when Kernel has custom names…

prophetz:
…pace

### DIFF
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -319,7 +319,7 @@ class Symfony extends Framework implements DoctrineProvider, PartedModule
         throw new ModuleRequireException(
             self::class,
             "Kernel class was not found.\n"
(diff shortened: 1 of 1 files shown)


src/Codeception/Module/Symfony.php
```diff
@@ -299,7 +299,7 @@ protected function getKernelClass(): string
             throw new ModuleRequireException(
                 self::class,
                 "File with Kernel class was not found at {$path}.\n"
-                . 'Specify directory where file with Kernel class for your application is located with `app_path` parameter.'
+                . 'Specify directory where file with Kernel class for your application is located with `kernel_class` parameter.'
```
TavoNiievez, line 302: it seems that in this specific message we are really waiting for the `app_path`.

In the message you corrected below it is correct to change to `kernel_class`.

TavoNiievez (changes requested):

TavoNiievez (approved):

PR #205, opened 2025-01-19
Added Symfony Translation assertions

TavoNiievez:
Examples:
```php
    $I->amOnPage('/register');
    $I->dontSeeFallbackTranslations();
    
    $I->amOnPage('/');
    $I->dontSeeMissingTranslations();
   
    $I->amOnPage('/register');
    $defined = $I->grabDefinedTranslationsCount();
    $I->assertSame($defined, 6);
   
    $I->amOnPage('/login');
    $I->seeAllTranslationsDefined();
    
    $I->amOnPage('/register');
    $I->seeDefaultLocaleIs('en');

    $I->amOnPage('/register');
    $I->seeFallbackLocalesAre(['es']);

    $I->amOnPage('/register');
    $I->seeFallbackTranslationsCountLessThan(1);

    $I->amOnPage('/');
    $I->seeMissingTranslationsCountLessThan(1);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
         "symfony/security-core": "^5.4 | ^6.4 | ^7.2",
         "symfony/security-csrf": "^5.4 | ^6.4 | ^7.2",
         "symfony/security-http": "^5.4 | ^6.4 | ^7.2",
+        "symfony/translation": "^5.4 | ^6.4 | ^7.2",
         "symfony/twig-bundle": "^5.4 | ^6.4 | ^7.2",
         "symfony/validator": "^5.4 | ^6.4 | ^7.2",
         "symfony/var-exporter": "^5.4 | ^6.4 | ^7.2",
--- a/src/Codeception/Module/Symfony.php
+++ b/src/Codeception/Module/Symfony.php
@@ -25,6 +25,7 @@ use Codeception\Module\Symfony\SecurityAssertionsTrait;
 use Codeception\Module\Symfony\ServicesAssertionsTrait;
 use Codeception\Module\Symfony\SessionAssertionsTrait;
 use Codeception\Module\Symfony\TimeAssertionsTrait;
+use Codeception\Module\Symfony\TranslationAssertionsTrait;
 use Codeception\Module\Symfony\TwigAssertionsTrait;
 use Codeception\Module\Symfony\ValidatorAssertionsTrait;
 use Codeception\TestInterface;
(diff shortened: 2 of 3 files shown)


ThomasLandauer:
I've never used translations, so I can't say much. Just one idea for the 3 `...CountLessThan()` methods:
* Why don't you assert the exact number? `...CountEquals()`?
* Or maybe just *return* the number with `grabTranslation...MessageCount()`, then have the user assert whatever they want?

xEdelweiss:
This looks like a great addition! I have a few suggestions:

The function names feel a bit verbose. What do you think about simplifying them to something like:
- `dontSeeMissingTranslations()`
- `dontSeeFallbackTranslations()`
- `dontSeeDefinedTranslations()`
- `seeMissingTranslationsCountLessThan()`
- `seeFallbackTranslationsCountLessThan()`
- `seeDefinedTranslationsCountLessThan()`

I do not know what Defined means in the context of Translations. But if it's a translation that is not fallback or missing, then it would be great to have `seeAllTranslationMessagesDefined()` (or `seeAllTranslationsDefined()`) as an alias for `dontSeeTranslationMissingMessages() + dontSeeTranslationFallbackMessages()`.

Silly question: shouldn't the translation locale be the same as the locale? :)

ThomasLandauer:
Yeah, the shorter names are indeed better IMO.
* But I would rename `grabDefinedTranslations()`: You're not grabbing the actual translations, so it should be something like `grabDefinedTranslationsCount` or `grabNumberOfDefinedTranslations`.
* About the `...TranslationsCountLessThan()` approach, after some more thinking about it: What I don't like is the overall notion that *some* errors are OK. To me, this sounds like seeElement*Sometimes*()... If Codeception had an HTML validator built in, I also wouldn't want it to be called `seeHtmlErrorsLessThan()`. To me, this is not what a testing framework should propagate. I didn't really doublecheck, but I think this would be the only method in Codeception which does something like an "estimation".

But again, I'm not using translations, and since you're the one who did the work, you're free to decide what you need/like best :-)

ThomasLandauer:
Ahem, HTML errors are deterministic too - if you adhere to the specs, you have 0 errors in each run. `seeRequestTimeIsLessThan` is an excellent example, cause nobody would ever want to assert `seeRequestTimeIsExactly(802)` ;-)
And exactly because the translations are deterministic, I think the `grab` approach is possible, since you'll always get the exact same number.

But it's not that important. You're the boss -> just merge it! :-)

ThomasLandauer:
See also https://github.com/Codeception/Codeception/discussions/6821

PR #206, opened 2025-02-17
Added Symfony Logger assertion (dontSeeDeprecations)

TavoNiievez:
```php
       $I->amOnPage('/home');
       $I->dontSeeDeprecations();
```

```cmd
 Step  Don't see deprecations
 Fail  Found 2 deprecation messages in the log:
  - Since doctrine/doctrine-bundle 2.12: The default value of "doctrine.orm.controller_resolver.auto_mapping" will be changed from `true` to `false`...
  - Since doctrine/doctrine-bundle 2.13: Enabling the controller resolver automapping feature has been deprecated....
Failed asserting that an array is empty.

```

### DIFF
--- a/src/Codeception/Module/Symfony/LoggerAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/LoggerAssertionsTrait.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Codeception\Module\Symfony;
+
+use Symfony\Component\HttpKernel\DataCollector\LoggerDataCollector;
+use Symfony\Component\VarDumper\Cloner\Data;
+use function sprintf;
+
+trait LoggerAssertionsTrait
+{
+    /**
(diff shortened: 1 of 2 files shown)


src/Codeception/Module/Symfony/LoggerAssertionsTrait.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Codeception\Module\Symfony;
+
+use Symfony\Component\HttpKernel\DataCollector\LoggerDataCollector;
+use Symfony\Component\VarDumper\Cloner\Data;
+use function sprintf;
+
+trait LoggerAssertionsTrait
+{
+    /**
+     * Asserts that there are no deprecation messages in the log.
```
ThomasLandauer, line 14: Asserts that there are no deprecation messages in Symfony's log.

ThomasLandauer (changes requested):

PR #210, opened 2025-05-21
Update ServicesAssertionsTrait.php: Adding another hint about private…

ThomasLandauer:
… services

If this setting is missing, `test.private_services_locator` isn't loaded. Just encountered this problem in one of my projects...


ThomasLandauer:
I reworded it completely.

PR #227, opened 2026-03-01
Allow execution via PHPUnit

TavoNiievez:
### 🚀 Added
- It is now possible to use the module only with PHPUnit through `CodeceptTestCase` while maintaining the standard Codeception workflow.
- Internal Symfony+PHPUnit test application under `tests/_app`.
- PHP CS Fixer configuration for consistent styling.

### 🔧 Changed
- Simplified CI workflow in `.github/workflows/main.yml` to match the current validation flow.
- Refactored and reduced complexity across `src/` for maintainability.
- Updated `composer.json` to support the new workflow.
- Added cache support as part of module capabilities.


ThomasLandauer:
Sorry, I can't say anything about the code :-(

It sounds like a very good improvement, but I don't understand if fully yet.
Right now, the Symfony module is talking to Codeception's "core", which somehow "converts" the tests and runs them through PHPUnit?!
So with this PR, the Symfony module is now talking to PHPUnit directly? But what do you mean by "maintaining the standard Codeception workflow" then?

ThomasLandauer:
Wow, even with diagram :-)

Let me raise some questions we could discuss, in order to explain/document this. (I've never used PHPUnit, so some are probably quite basic.)
1. This is a real game-changer, isn't it? Now it's not only *Codeception's* Symfony module, but a Symfony plugin/addon/module for PHPUnit and Codeception? Have you thought about giving it a new name?
2. So it's a *replacement* for Symfony's built-in `WebTestCase`? What are the advantages? Would it be possible to create a comparison table?
3. If you use it in PHPUnit, you certainly can't access other Codeception modules (Doctrine, REST, Asserts, ...)? So what's the plan for those who need this?
4. Do other "extensions" like this exist for PHPUnit?

TavoNiievez:
@ThomasLandauer Those are all excellent questions...

1. You're right that this makes the code more than just a module for Codeception. It could now be a standalone plugin/add-on, and:
   * I'm not sure if it's best to keep it in the same repository.
   * It's confusing because the Codeception module-symfony recipe will still trigger during installation.

2. For now, I’ve decided not to make `CodeceptTestCase` public and to use it only internally, so I can keep working on this while I figure out a better answer for you.

3. Indeed, this will only work for the methods in this module and not for all Codeception methods; however, in many cases, Codeception methods are merely syntax sugar, and it is highly likely that the user already has an alternative they can use in their application.

4. Off the top of my head, https://github.com/zenstruck/mailer-test comes to mind, although the syntax it uses is more Laravel-style, and it's only for the mailer component.

I don't want to delay the rest of the changes in this PR any further, so I'll release it as is. We can continue discussing whether it's better to make CodeceptTestCase public in this same repository or not, and how to resolve the issue with the recipe.

PR #232, opened 2026-05-22
Support RawMessage instead of Email for Mail/Mime assertions

### LINKED ISSUE #233: Mailer and Mime assertion trait cannot handle encrypted emails
Heyho,

while writing some functional tests for sending encrypted and signed emails in our application I saw that the [MailerAssertionsTrait](https://github.com/Codeception/module-symfony/blob/main/src/Codeception/Module/Symfony/MailerAssertionsTrait.php) and the [MimeAssertionTrait](https://github.com/Codeception/module-symfony/blob/main/src/Codeception/Module/Symfony/MimeAssertionsTrait.php) are using Symfony's [Email](https://github.com/symfony/mime/blob/8.1/Email.php) class rather than the parent [Message](https://github.com/symfony/mime/blob/8.1/Message.php) or even [RawMessage](https://github.com/symfony/mime/blob/8.1/RawMessage.php) classes.

Symfony also provides a [MailerAssertionsTrait](https://github.com/symfony/framework-bundle/blob/8.1/Test/MailerAssertionsTrait.php) in the Framework Bundle which uses RawMessage everywhere.

The issue now is the following:
Symfony's [SMimeEncrypter](https://github.com/symfony/mime/blob/8.1/Crypto/SMimeEncrypter.php#L44) class does not return an Email object but a Message object instead when used like [documented](https://symfony.com/doc/current/mailer.html#encrypting-messages).

Codeception's [MailerAssertionsTrait::grabLastSentEmail](https://github.com/Codeception/module-symfony/blob/main/src/Codeception/Module/Symfony/MailerAssertionsTrait.php#L103) has `?Email` as return type which will cause a type error here since the [MessageEvent](https://github.com/symfony/mailer/blob/8.1/Event/MessageEvent.php#L40) fetched from Symfony's [MessageLoggerListener](https://github.com/symfony/mailer/blob/8.1/EventListener/MessageLoggerListener.php#L43) contains a `Message` Object and not an object of its child class `Email`.

I already prepared a fix for this here:
https://github.com/Codeception/module-symfony/pull/232

There's just the question whether to use `RawMessage` as type hint everywhere, like Symfony does in its own MailerAssertionsTrait or whether to use the "biggest" supported type for each of the [Mime Constraints](https://github.com/symfony/mime/tree/8.1/Test/Constraint) used (e.g. `Message` for [EmailHasHeader](https://github.com/symfony/mime/blob/8.1/Test/Constraint/EmailHasHeader.php#L34) constrait, as it does not support `RawMessage`) like I did in my current fix.

I already prepared the tests in the symfony-module-tests repo here: https://github.com/Codeception/symfony-module-tests/pull/40

Thanks for help and your great work =)

krrico:
Adjusted type hints of Mailer and Mime assertions traits to support RawMessage and Message instead of Email only in order to also support fetching sent Emails when they were encrypted or signed before which turns them into plain Message objects rather than Email objects (which is a child class of Message).

### DIFF
--- a/src/Codeception/Module/Symfony/MimeAssertionsTrait.php
+++ b/src/Codeception/Module/Symfony/MimeAssertionsTrait.php
@@ -16,9 +16,9 @@ use function sprintf;
 trait MimeAssertionsTrait
 {
     /**
-     * Verify that an email contains addresses with a [header](https://datatracker.ietf.org/doc/html/rfc4021)
+     * Verify that a message contains addresses with a [header](https://datatracker.ietf.org/doc/html/rfc4021)
      * `$headerName` and its expected value `$expectedValue`.
-     * If the Email object is not specified, the last email sent is used instead.
+     * If no Message is specified, the last sent message is used instead.
      *
      * ```php
      * <?php
@@ -27,13 +27,12 @@ trait MimeAssertionsTrait
      */
     public function assertEmailAddressContains(string $headerName, string $expectedValue, ?Message $email = null): void
     {
-        $email = $this->verifyEmailObject($email, __FUNCTION__);
-        $this->assertThat($email, new MimeConstraint\EmailAddressContains($headerName, $expectedValue));
+        $this->assertThat($this->getMessageOrFail($email, __FUNCTION__), new MimeConstraint\EmailAddressContains($headerName, $expectedValue));
     }
 
     /**
-     * Verify that an email has sent the specified number `$count` of attachments.
-     * If the Email object is not specified, the last email sent is used instead.
+     * Verify that an email has the specified number `$count` of attachments.
+     * If no Email is specified, the last sent email is used instead.
      *
      * ```php
      * <?php
@@ -42,13 +41,12 @@ trait MimeAssertionsTrait
      */
     public function assertEmailAttachmentCount(int $count, ?Email $email = null): void
     {
-        $email = $this->verifyEmailObject($email, __FUNCTION__);
-        $this->assertThat($email, new MimeConstraint\EmailAttachmentCount($count));
+        $this->assertThat($this->getMessageOrFail($email, __FUNCTION__), new MimeConstraint\EmailAttachmentCount($count));
     }
 
     /**
-     * Verify that an email has a [header](https://datatracker.ietf.org/doc/html/rfc4021) `$headerName`.
-     * If the Email object is not specified, the last email sent is used instead.
+     * Verify that a message has a [header](https://datatracker.ietf.org/doc/html/rfc4021) `$headerName`.
+     * If no Message is specified, the last sent message is used instead.
      *
      * ```php
      * <?php
@@ -57,14 +55,13 @@ trait MimeAssertionsTrait
      */
     public function assertEmailHasHeader(string $headerName, ?Message $email = null): void
     {
-        $email = $this->verifyEmailObject($email, __FUNCTION__);
-        $this->assertThat($email, new MimeConstraint\EmailHasHeader($headerName));
+        $this->assertThat($this->getMessageOrFail($email, __FUNCTION__), new MimeConstraint\EmailHasHeader($headerName));
     }
 
     /**
      * Verify that the [header](https://datatracker.ietf.org/doc/html/rfc4021)
-     * `$headerName` of an email is not the expected one `$expectedValue`.
-     * If the Email object is not specified, the last email sent is used instead.
+     * `$headerName` of a message is not the expected one `$expectedValue`.
+     * If no Message is specified, the last sent message is used instead.
      *
      * ```php
      * <?php
@@ -73,14 +70,13 @@ trait MimeAssertionsTrait
      */
     public function assertEmailHeaderNotSame(string $headerName, string $expectedValue, ?Message $email = null): void
     {
-        $email = $this->verifyEmailObject($email, __FUNCTION__);
-        $this->assertThat($email, new LogicalNot(new MimeConstraint\EmailHeaderSame($headerName, $expectedValue)));
+        $this->assertThat($this->getMessageOrFail($email, __FUNCTION__), new LogicalNot(new MimeConstraint\EmailHeaderSame($headerName, $expectedValue)));
     }
 
     /**
      * Verify that the [header](https://datatracker.ietf.org/doc/html/rfc4021)
-     * `$headerName` of an email is the same as expected `$expectedValue`.
-     * If the Email object is not specified, the last email sent is used instead.
+     * `$headerName` of a message is the same as expected `$expectedValue`.
+     * If no Message is specified, the last sent message is used instead.
      *
      * ```php
      * <?php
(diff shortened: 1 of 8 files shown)


krrico:
Just added the Test PR here:
https://github.com/Codeception/symfony-module-tests/pull/40

src/Codeception/Module/Symfony/MimeAssertionsTrait.php
```diff
@@ -85,7 +87,7 @@ public function assertEmailHeaderNotSame(string $headerName, string $expectedVal
      * $I->assertEmailHeaderSame('To', 'jane_doe@gmail.com');
      * ```
      */
-    public function assertEmailHeaderSame(string $headerName, string $expectedValue, ?Email $email = null): void
+    public function assertEmailHeaderSame(string $headerName, string $expectedValue, ?Message $email = null): void
```
krrico, line 86: I'm not sure whether it's better to use RawMessage as type hint everywhere and let the Symfony Mime Constraints used here run into the RuntimeException when they don't support RawMessage or Message.

TavoNiievez:
Hey @krrico , thanks for the thorough bug report and the PR — you identified the root cause precisely and your proposed fix was on the right track. 🙌

I went ahead and implemented a refined version of your approach in [e8ab243](https://github.com/Codeception/module-symfony/commit/e8ab243). Here's a summary of the decisions made and why.

## The core issue you correctly identified

When Symfony's `SMimeEncrypter` (or `SMimeSigner`) wraps an `Email` before sending, the result stored in the `MessageLoggerListener` is a plain `Message` — not an `Email`. The original `grabLastSentEmail()` returned `?Email`, causing a type error, and all `MimeAssertionsTrait` methods expected `?Email`, making them unusable in that scenario.

## Why we didn't use `RawMessage` everywhere

Your PR took the straightforward approach of widening every type hint to `RawMessage` (mirroring what Symfony's FrameworkBundle `MailerAssertionsTrait` does). That works, but it pushes errors further down the stack: if a `RawMessage` (the bare base class) is passed to a Mime constraint that requires at least a `Message`, Symfony throws a `LogicException` at runtime — without a helpful message.

I chose to use the **semantically correct minimum type for each assertion** instead, which makes type errors explicit at the PHP level and produces better error messages:

| Assertion | Parameter type | Why |
|-----------|---------------|-----|
| `assertEmailHasHeader` | `?Message` | `EmailHasHeader` constraint requires `Message` (rejects `RawMessage`) |
| `assertEmailAddressContains` | `?Message` | `EmailAddressContains` constraint requires `Message` |
| `assertEmailHeaderSame` | `?Message` | `EmailHeaderSame` constraint requires `Message` |
| `assertEmailHeaderNotSame` | `?Message` | `EmailHeaderSame` constraint requires `Message` |
| `assertEmailNotHasHeader` | `?Message` | `EmailHasHeader` constraint requires `Message` |
| `assertEmailAttachmentCount` | `?Email` | `EmailAttachmentCount` constraint requires `Email` specifically |
| `assertEmailHtmlBodyContains` | `?Email` | `EmailHtmlBodyContains` constraint requires `Email` specifically |
| `assertEmailHtmlBodyNotContains` | `?Email` | `EmailHtmlBodyContains` constraint requires `Email` specifically |
| `assertEmailTextBodyContains` | `?Email` | `EmailTextBodyContains` constraint requires `Email` specifically |
| `assertEmailTextBodyNotContains` | `?Email` | `EmailTextBodyContains` constraint requires `Email` specifically |

The distinction maps directly to what Symfony's own Mime constraints check at runtime: header-based assertions only need a `Message` (which carries headers), while body and attachment assertions need a full `Email` (which adds body and part accessors). Since encrypted/signed messages are still `Message` instances — they have headers but no plain-text body — the header-based assertions will work on them out of the box.

## The internal method: `verifyEmailObject` → `getMessageOrFail`

Your PR widened `verifyEmailObject` to `(?RawMessage): RawMessage` to match the broadest possible type. I renamed it to `getMessageOrFail` and gave it a tighter contract:

```php
private function getMessageOrFail(?Message $message, string $function): Message
```

- **Accepts `?Message`** — so it is compatible with all public callers.
- **Returns `Message`** — the minimum type all Mime constraints accept.
- **Fails explicitly on bare `RawMessage`** — if neither a message is provided nor one was sent, or if what was sent is a plain `RawMessage` with no headers, it fails the test with a clear message rather than letting Symfony's constraint throw a `LogicException` deep inside the assertion.

## What `grabLastSentEmail()` returns now

`grabLastSentEmail()` now returns `?RawMessage` (via an internal `grabLastSentRawMessage()` helper), matching the raw output of `MessageLoggerListener`. This means it covers both `Email` and `Message` objects. If you need to type-narrow the result yourself, `instanceof` is the right tool:

```php
$message = $I->grabLastSentEmail(); // ?RawMessage — works for Email and Message
if ($message instanceof Email) {
    // access Email-specific methods
}
```

## Tests added

I also added tests that cover your exact use case — sending a plain `Message` through the mailer and asserting on its headers:

```php
// In MimeAssertionsTest
public function testHeaderAssertionsWorkWithSentMessage(): void
{
    // /send-message dispatches a Message (not Email) through the mailer
    $this->client->request('GET', '/send-message');

    $this->assertEmailHasHeader('To');
    $this->assertEmailAddressContains('To', 'jane_doe@example.com');
    $this->assertEmailHeaderSame('Subject', 'Test message');
}
```

## One thing I'd like to ask of you

Could you update your PR in the testing repository ([Codeception/symfony-module-tests#40](https://github.com/Codeception/symfony-module-tests/pull/40)) to match the simplified approach I've taken here?

Your current commit ([71d207e](https://github.com/Codeception/symfony-module-tests/commit/71d207ed9baddde8a9f38f565497d8c74ae300f)) uses actual S/MIME encryption — a `Crypto` helper that generates real OpenSSL certificates at runtime, an `ext-openssl` dev dependency, and a `SendEncryptedEmailController` that calls `SMimeEncrypter::encrypt()`. That's a faithful reproduction of the real-world scenario, but it introduces a non-trivial infrastructure burden on the test suite (an OpenSSL extension requirement, ephemeral certificate files, etc.).

The key insight is that for testing the module's assertion methods, **we don't actually need to encrypt anything**. What matters is that the mailer receives a `Message` object instead of an `Email`. The simplest way to achieve that is to just construct and send a `Message` directly — no encryption, no certificates, no OpenSSL.

The simplified version I've added here does exactly that:

```php
// tests/_app/Mailer/MessageMailer.php
public function send(string $recipient): void
{
    $headers = new Headers();
    $headers->addMailboxListHeader('From', ['no-reply@example.com']);
    $headers->addMailboxListHeader('To', [$recipient]);
    $headers->addTextHeader('Subject', 'Test message');

    $this->mailer->send(new Message($headers, new TextPart('Message body content')));
}
```

So in the testing repo, the ideal update would be to:
- Replace the `SendEncryptedEmailController` / `Crypto` / `SMimeEncrypter` setup with a simple controller that sends a plain `Message`
- Remove the `ext-openssl` dev dependency from `composer.json`
- Update `IssuesCest` to assert against this simpler `/send-message` endpoint

Thanks again for the detailed issue and the groundwork in the PR!

TavoNiievez (approved):

src/Codeception/Module/Symfony/MimeAssertionsTrait.php
krrico: Resolving this conversation as we keep going with supporting the minimum version to prevent running into RuntimeExceptions as mentioned here:
https://github.com/Codeception/module-symfony/pull/232#issuecomment-4634004614

krrico:
Thanks a lot @TavoNiievez for jumping in and pushing this forward =)

As mentioned [here already](https://github.com/Codeception/module-symfony/pull/232#discussion_r3288574506) I also liked to go with using the minimum supported version to prevent to run into the RuntimeExceptions, so your thoughts and changes here perfectly match <3

I fixed and force pushed my changes in the [Test PR](https://github.com/Codeception/symfony-module-tests/pull/40) and also had some doubts about bringing in then OpenSSL dependency as [mentiond here](https://github.com/Codeception/symfony-module-tests/pull/40#discussion_r3288588326).

So the Test PR is way smaller and easier to review now and still runs fine locally after copying the changes from here to the vendor directory in the test repo.

